PR #10, opened 2015-10-29
GStreamer-base: update riff fourcc patch

athoik:
More info: http://forums.openpli.org/topic/29501-gstreamer-10/page-92


eriksl:
This patch mixes functional and cosmetic. Please split.

PR #11, opened 2015-10-30
GStreamer-base: add patch for ffmpeg mpeg4 riff media

athoik:
More info: http://forums.openpli.org/topic/29501-gstreamer-10/page-92


eriksl:
Applied.

PR #14, opened 2015-11-03
Disable certificate verification by default in python 2.7.9

Taapat:
Sertificate verification by default enbled in python 2.7.9 in PEP 0476, but it breaks many things where used urllib2. Therefore disable it. But this still allows the use of certificate verification using the context if necessary.


eriksl:
Is there a related discussion on the forum?

Taapat:
I started the discussion here: http://forums.openpli.org/topic/38580-gstreamer-1x-images-for-testing/page-13#entry511057
But there is not a big response.

milosoftware:
Disabling certificate verification in Enigma2 is something one can live with, after all, streaming data over HTTPS is just plain stupid.

This patch will globally disable all verification in the Python base libraries, now that will lead to some unexpected results.

(Please keep in mind that HTTPS without certificate verification is pointless, it adds no security at all compared to plain HTTP. It's the certificates that really matter, disabling encryption would offer more security than skipping verification.)

If this means we have to adjust dozens of plugins, well, I'll happily push those for you.

PR #16, opened 2015-11-07
Addtranslation files in xmltvimport plugin package

Taapat:
This is a little hack, which I personally don't like, but I see that there are already other hacks for plugin.png and egg-info.


milosoftware:
Hacks are to be fixed, not invitations to add more.
Adjust the installer of xmltvimport to install the files you want.

Taapat:
I agree that the hacks need to fix.
Maybe I'm wrong, but it seems to me that in this case it is necessary to also modify xmltvimport plugin, so I chose the simplest way as a temporary solution until someone changes this all.
If you do not like it, ok, we will continue to build plugin without translation files.

PR #17, opened 2015-11-08
Remove dreamcrc library hack in plugin xmltvimport

Taapat:
It must be fixed in https://github.com/OpenPLi/enigma2-plugin-extensions-xmltvimport/pull/7
After add this, please do not forget update the plugin version sha to actual.

The next my step will be to simplify bitbake file using distutils-openplugins and remove all hacks. But I will wait what do you say about this commit.


milosoftware:
Merged.

PR #29, opened 2015-12-31
CDtextinfo: It's not installed by default.Bring it back.This partiall…

Hains:
…y revert commit 22f689ff821bf2a19832448a96c083225cef7461.


Hains:
No need for that..audio and DVD plays fine(internal drive).

Hains:
The modules are already installed by default...so just cdtextinfo.

PR #31, opened 2016-01-10
Remove sh from busybox.links.nosuid

Taapat:
This fix busybox package update.
It looks like the trick with sh copying in tmp does not work.
update-alternatives is bash script that link to /bin/sh, and after update-alternatives remove /bin/sh, it stops working.
Link to /bin/sh already exists in the package, there is no need to create it, if remove it from the busybox.links.nosuid, it is not used in update-alternatives and not removed when run update-alternatives --remove.


milosoftware:
Looks like something that should be sent upstream, instead of just putting a workaround in our system.

Taapat:
After this fix busybox upgrade without problems.
I assume that this fix is also required in the master not only in master-next.
But I building branch master-next and I can't test master.
I do not know how busybox package work in another system, perhaps there is used a different method of installation, so I suggest fix here, where I can test them.

Taapat:
And I not change busybox version number, so ipk package should not automatically update only rebuild.

PR #32, opened 2016-01-12
samba4: on install add directory /run/samba to fix building

Taapat:
To fix startup I move sockets-dir from /run/samba to /run.
Perhaps therefore do_install now wants to remove the directory /run/samba which no longer exists.
I added an empty directory to avoid error.
Perhaps reason is a another changes on openembedded, but in any case this fix samba building.


milosoftware:
There are patches pending in OE (master-next) that fix that issue. And some more (like moving a few files around).
They weren't merged as quickly as I hoped, so I'll just take this patch and merge what mainline already has done (enabling parallel build for example).

PR #35, opened 2016-01-24
enigma2-plugins: change only python plugins arch to all

Taapat:
If in the plugin is only python code, why it architecture depends of the machine?
So change only python plugins architecture to all.


milosoftware:
Squashed into a single commit and merged

PR #38, opened 2016-02-17
Vsftpd

Taapat:
**1. Set EXTRA_OEMAKE for vsftpd**
Default value for EXTRA_OEMAKE on openembedded-core removed in https://github.com/openembedded/openembedded-core/commit/aeb653861a0ec39ea7a014c0622980edcbf653fa
This break vsftpd building.
Therefore set the EXTRA_OEMAKE in vsftpd bb file.
In the meantime, while you're not update OE, problems with EXTRA_OEMAKE in vsftpd not exist.
It will be only after the OE update.
But it's not break anything, so maybe better to fix it now, so after OE update vsftpd building without error.

**2. Update vsftpd to 3.0.3**
Maybe you can update vsftpd also in the master, but I think that at this moment it is better to do so only in master-nex.


Taapat:
Sorry, something is wrong in this my pull request

PR #60, opened 2016-03-19
Fixpath to python-lxml install directory

Taapat:


Taapat:
If you use cherry-pick, why you not close pull request?

PR #69, opened 2016-04-26
[mdev-mount.sh] add mount option noatime for all type filesystem

Dima73:
discussion here
http://forums.openpli.org/topic/41278-ssd-trim/?view=findpost&p=544398


Taapat:
Maybe I am missing something, but why is it necessary variable FLAG?
Why not simply:
f ! mount -t auto -o noatime /dev/$MDEV "${MOUNTPOINT}" ; then

Dima73:
This variable is an example for the future.
That's clearer.

littlesat:
Can you please explain why this is required?
In addition I suggest we do not need that separete flag stuff.... when done it can simply be added as is.

Dima73:
This option is consistent with the Milo

PR #74, opened 2016-05-23
feed switch for OpenPLi Makefile

persianpros:
It's easier to have this switch instead of using bitbake and extra commands.


persianpros:
You can delete this as I sent a new one here : https://github.com/OpenPLi/openpli-oe-core/pull/81

PR #85, opened 2016-06-21
meta-dream: update dm7020hd drivers

athoik:
- fixed support for more than two simultaneously used tuners
- improved support for transponders with higher datarates via CI


eriksl:
Applied.

PR #93, opened 2016-08-26
python-mutagen: change SRC_URI to a valid one location

athoik:
Since googlecode was gone.

Also fix warnings when bulding OWIF due to mix of spaces and tabs introduces on a8fc9b9


athoik:
See if using  SRC_URI[md5sum] = "0c2cd954e4bacd79fadd45afc4acce4c" works.

PR #94, opened 2016-08-27
add USB DVB driver for Vuplus Tuner Turbo

Dima73:


Taapat:
Why allarch?
Or for all receivers is available vuplus-tuner-turbo driver?
I think that this package is intended only for vuplus receivers.

Dima73:
How to fix it?

athoik:
I think all "meta" packages should be inherit allarch.

Description should change and include for Vu+ only.

Dima73:
I do not see any problems.
For other models, not Vu+, just do not install.

Taapat:
I do not agree that "meta" packages should be inherit allarch.

I think that on feed should be only packages whose working.
If this package only works on a particular receiver, then it must be on the receiver of feed, not on all.

I think that needs change inherit allarch to PACKAGE_ARCH = "${MACHINE_ARCH}"

Taapat:
And vuplus-tuner-turbo license is "CLOSED",
why require conf/license/license-gplv2.inc?

RobvanderDoes:
> I think that on feed should be only packages whose working.
> If this package only works on a particular receiver, then it must be on the receiver feed, not on all.

Agreed, but that should apply to all plugins. PLi has many hardware-specific plugins on the feeds for all boxes.

athoik:
Sleep it simple guys, one ipk instead of several ipk, simple bitbake instead of complicated one.

Again this bitbake has to be in all arch, there is no machine code in. Just a dummy package to recommend few others. Even if you find it on non Vu+ boxes, installation will be fine.

RobvanderDoes:
> Even if you install it on non Vu+ boxes, installation will be fine.

True.

eriksl:
Why close?

The only thing that needs to be done is that the arch is changed to machine_arch and the plugin is only built for VU+ machines. That is accomplished by adding it to the BSP, create a pull request for VU+.

RobvanderDoes:
Why only for VU+ machines? The tuner works perfect on other brands as well.

Taapat:
Maybe I'm wrong, but I have seen vuplus-tuner-turbo only on vuplus BSP.
Maybe other brands no need for a separate driver.
In this case, the package may remain such as is now.

But really tuner works perfect on all other brands?

RobvanderDoes:
Provided they have the same kernel version (but that's the case with all drivers).
Anyway: there's nothing brand-specific in the driver-package.

> I have seen vuplus-tuner-turbo only on vuplus BSP.

Probably because VU maintains that repos.

Taapat:
But vuplus-tuner-turbo package is with the machine architecture.
You can not install from their from feed without forced options.

Taapat:
I think that if tuner working on other machines, then on their kernel is already included all the necessary and additional driver from vuplus not needed.

Dima73:
''I think that needs change inherit allarch to PACKAGE_ARCH = "${MACHINE_ARCH}"
and add check in openpli-enigma2-feed.bb to build only for vuplus.''

How to do it?

Dima73:
${@base_contains("MACHINE_FEATURES", "vuplus", "enigma2-plugin-drivers-dvb-usb-turbo" , "", d)}
???

Taapat:
I think:
${@bb.utils.contains('EXTRA_IMAGEDEPENDS', 'vuplus-tuner-turbo', 'enigma2-plugin-drivers-dvb-usb-turbo', '', d)}

eriksl:
So apparently the crux is whether the turbo tuners are meant only for VU+ receiver or for all receivers. I am inclined to think the first and in that case all the code should be in VU+ BSP. If not, we should make a package ourselves, I guess, that contain the necessary drivers and include them for every mipsel based receiver.

PR #101, opened 2016-09-11
Add unjffs2 tool

Dima73:


littlesat:
Why do we need this?

Dima73:
Only dreambox reciever.
But it seems not to work, for example 800se.

PR #105, opened 2016-09-12
Add astra-sm and dvblast utilities

athoik:
Astra (Advanced Streamer) is a professional software to organize Digital TV Service for TV operators and broadcasters, internet service providers, hotels, etc.

DVBlast is written to be the core of a custom IRD, CID, or ASI gateway, based on a PC with a Linux-supported card. It is very lightweight and stable, designed for 24/7 operation.


littlesat:
Can you please describe more what this means?

athoik:
Those are dvb utilies to perform several tasks our of enigma2 scope.

Eg with astra-sm you can use t2-mi or convert multicast streams to unicast or analyze TS streams.

On the other hand dvblast can stream a full transponder to your network.

See: http://www.videolan.org/projects/dvblast.html , https://cesbo.com/en/astra/quick-start/

PR #117, opened 2016-10-20
meta-openpli: fix python-mutagen md5 checksum failure on download

tomparkin:
The python-mutagen recipe set SRC_URI[md5sum] using a variable which doesn't
seem to be correctly evaluated, resulting in this kind of error message
during source fetching:

ERROR: Fetcher failure for URL: <blah...>
Checksum mismatch!
'/home/foobar/openpli-oe-core/sources/mutagen-1.18.tar.gz' has md5
checksum 0c2cd954e4bacd79fadd45afc4acce4c when ${MD5SUM} was expected

This breaks the vuduo2 build for me.  Fix by expressing the md5sum as a
string literal where it is used.

### DIFF
--- a/meta-openpli/recipes-devtools/python/python-mutagen_1.18.bb
+++ b/meta-openpli/recipes-devtools/python/python-mutagen_1.18.bb
@@ -8,13 +8,11 @@ PR = "r1"
 DEPENDS = "python"
 RDEPENDS_${PN} = "python-shell"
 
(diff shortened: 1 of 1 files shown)


WanWizard:
@athoik can you verify this? Since you've changed this in August, if this is an issue, why wasn't it spotted earlier?

athoik:
Most probably I was checking in OpenPLi 5, or I didn't clean the package, so it was always in my sources folder.

So it's fine to commit.

PR #119, opened 2016-10-21
add local openpli file umountfs

Dima73:
-sleep 5 sec. epg time to save
-sync to flush pending writes for loop-mounted file system
-add 'sdparm --command=stop /dev/sd?' for sleep hdd


Taapat:
Do we really need to add a total delay of 10 seconds when switch off or a complete restart reciver to save epg?
How big is this file if it stored so long?

Dima73:
5 sec. for sleep hdd need.Tested my usb hdd,need around 4.5 sec.

Taapat:
But what if someone has a slower disk?
And what about those who do not use very large epg files?

Perhaps it is possible to obtain real information when this file is saved?
I do not think it was right to blindly introduce a pause, which probably is longer that the total receiver switch off time.

Taapat:
And maybe before add some delay, it is necessary to check whether such devices are used?

PR #122, opened 2016-10-24
Update libbluray

Taapat:
Update on latest git revision.
Enable udf (as submodule) to read iso discs.
Add patch to expose clip_id in BLURAY_CLIP_INFO. This allow get m2ts track name from BLURAY_CLIP_INFO in BlurayPlayer plugin.


Taapat:
Warning!
To change the git protocol to use submodules is necessary to clean the previous git.
I do not know how to work in your build system.
Maybe I need to change the package name?

PR #140, opened 2016-12-04
dvbsnoop: Use "AUTOREV" for future updates

persianpros:


eriksl:
Requests for AUTOREV are not honoured.

PR #145, opened 2016-12-11
[mdev-mount] blacklisted uno4k/solo4k/ultimo4k/hd51 internal mmc

Dima73:


littlesat:
Why?

Dima73:
Internal mmc is flash and mounted as /usb .
We do not need this.

littlesat:
Why?

Dima73:
Flash automatic  mounted as /media/usb, why?

WanWizard:
Seems logical to me. Flash is already mounted, you don't need to mount it again, and if you want to mount it, I assume you mount it as /media/mmc, and not as /media/usb.

Dima73:
/dev/root is link to mmcblk0p4
`root@vusolo4k:~# readlink /dev/root
mmcblk0p4
`

Dima73:
On new UHD receivers internal flash type is  eMMC instead of nand.
Therefore, for the root system it is mounted /dev/root, which is a link to mmcblk0p4.
 
In mdev-mount.sh provides that mmcblk0p1 is mounted as mmc1.
For mmcblk0p4 there not are any special conditions, therefore by default it is mounted as a USB.
 
I think that the root system do not need in addition mount in media.
This change does not allow mount the device if on it is a link from /dev/root and thefore the root system (in this case mmcblk0p4) are not addition mount in media.

PR #146, opened 2016-12-14
dvbsnoop : 1.4.52 with old DVB API fix

persianpros:
Now developers could build openpli for old boxes like dm800 too.


Taapat:
It is not necessary on openpli and with this you break multitream on all receivers who use the dvb api 5.
I explain why here: https://github.com/oe-alliance/oe-alliance-core/commit/4cc96c0428d9ff6859925fa18049db0fc125f38b

Taapat:
Try: https://github.com/persianpros/dvbsnoop/pull/4

athoik:
Please close that merge. Come back when checks are done properly.

PR #149, opened 2016-12-20
[mdev-mount.sh] first check this is mmc

Dima73:


Taapat:
Why?
I created a universal solution regardless of the filesystem.
Why do you want again use only specific filesystem to use it for specific receivers?

Dima73:
Why do we need further action, if it is not mmc?

Taapat:
Therefore, it is possible tomorrow will be another box that will use a different filesystem.
What is the problem with existing solutions?
Maybe I miss something, but I have not seen any complaints that **exactly these changes** affect anything.

PR #158, opened 2017-01-13
zgemma: update bsp

Taapat:
h3: move plugin vfdcontrol from image to feed

Please use together with this: https://github.com/OpenPLi/enigma2/pull/607

### DIFF
--- a/meta-openpli/recipes-openpli/enigma2/enigma2.bb
+++ b/meta-openpli/recipes-openpli/enigma2/enigma2.bb
@@ -183,6 +183,7 @@ EXTRA_OECONF = "\
 	--enable-dependency-tracking \
 	ac_cv_prog_c_openmp=-fopenmp \
 	${@base_contains("MACHINE_FEATURES", "textlcd", "--with-textlcd" , "", d)} \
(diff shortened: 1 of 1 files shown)


zgemma-star:
please add in meta-openpli/recipes-openpli/enigma2/enigma2.bb 
EXTRA_OECONF = "\
line: ${@base_contains("MACHINE_FEATURES", "7segment", "--with-7segment" , "", d)} \

Taapat:
For compatibility with other bsp would be better to use 7seg which is currently used to install a enigma2-plugin-systemplugins-vfdcontrol.
But I did not want to do this at the moment until I satisfied that my skin_text.xml working on all of these receivers.
But there is no problem in this moment add 7segment, if you want to use it, and later also add 7seg.

zgemma-star:
Actually we want to install as few as possible plugins from BSP and only define machine features.

I prefer to use as MACHINE_FEATURES ="7segment" but if you prefer we can change it in 7seg but this is now used by OPENPLI_FEATURES. With your great solution for the 7segment display problem we will remove the VFD plugin entire from BSP.

I hope you can specify the 7segment or 7seg into enigma2.bb so that the new 7segment display skin will been add in the new builds.

Taapat:
Ok I added 7segment on this moment.

PR #168, opened 2017-02-04
dmm: Fix non bootable box.

Hains:
Since commit:
http://git.openembedded.org/openembedded-core/commit/meta/classes/kernel.bbclass?h=morty&id=a248ef51ae680e81cf78f07fe242ac6e01a5fcb4
vmlinux is not incluced in image.
Which result in a non bootable box after a flash.
Set files to vmlinux to restore it.


Hains:
Perhaps the comment is not well, but feel free to change it or report me.

PR #173, opened 2017-02-25
[Samba] Update to 3.6.25

Schimmelreiter:
- No space gain thanks to OpenWrt patches removing unneeded features
- SMBv2 support as in Windows Vista rather than max. SMBv1 as in Windows 2000
  SMBv1 -> SMBv2 was the biggest update to SMB ever, see https://en.wikipedia.org/wiki/Server_Message_Block#SMB_2.0
- Full IPv6 compatibility
- Some defaults patched to make it a drop-in-replacement for Samba 3.0.37 (No config changes needed)

Also disabled the busybox mount.cifs so that cifs-utils can step in.


littlesat:
I'm afraid we stay on samba 4... and do not revert back to 3

Schimmelreiter:
That was for OpenPLi 4.0, it had Samba 3.0.x :)

WanWizard:
Let's keep this open for the moment, so we can decide later what to do with it. We need to keep OpenPLi 4 for some boxes we can't/won't build a newer version for,

PR #177, opened 2017-03-08
Add support for ch340/ch341 device

pauluzs:


Dima73:
For many receivers  config not enabled 
# CONFIG_USB_SERIAL_CH341 is not set

Dima73:
need add 
RRECOMMENDS_${PN} = "kernel-module-ch341"

PR #185, opened 2017-03-12
kodi: add recipe for osmega

athoik:
Since we are using the same BSP layer for OpenPLi 4 and OpenPLi Next it's hard to add the recipe only to BSP, beucase it will fail when Building for OpenPLi 4. So move the recipe into kodi folder.


WanWizard:
You might want to remove this, and re-add it without Yoo's email address?

athoik:
The same patch commited on OE-A as well. 

https://github.com/oe-alliance/oe-alliance-core/blob/4.0/meta-brands/meta-xcore/recipes-kodi/kodi/kodi-17.0/0001-Krypton-V3D-support.patch

PR #213, opened 2017-04-30
softcams: silence already stripped warning

athoik:


eriksl:
Applied (dev & master4).

PR #214, opened 2017-04-30
firewall: remove depends to iptables and virtual/kernel

athoik:
We don't really depend on those packages to build an allarch package.
The runtime depends will add required iptables and kernel modules.


eriksl:
Applied in dev.

PR #222, opened 2017-05-03
[satipclient] goto oe-alliance version

Dima73:


meta-openpli/recipes-openpli/satip/satipclient.bb
```diff
@@ -1,20 +1,32 @@
...
+INITSCRIPT_PARAMS = "defaults"
+
+do_install_append() {
+    install -d ${D}/etc/init.d
+    install -m 0755 ${WORKDIR}/satipclient.sh ${D}/etc/init.d/satipclient
 }
+
+EXTRA_OECONF = " \
+    --with-boxtype=${MACHINE} \
+    --with-machinebuild="${MACHINEBUILD}" \
```
athoik: We do not have MACHINEBUILD in our OE, do we need it?

Dima73: athoik
Do you want to delete this?

PR #227, opened 2017-05-04
Fix some URLs

MastaG:
This will make the develop branch of openpli-oe-core build openpli-enigma2-feed again from scratch.
- Update libev to 4.24 (4.23 doesn't exist anymore on the server)
- Update w_scan to 20170107 (20161022 doesn't exist anymore on the server)
- Update the URL for pyload-src-v0.4.9.zip. (github doesn't host it anymore for some reason so let's use the openelec.tv mirror)

### DIFF
--- a/meta-openpli/recipes-connectivity/pyload/pyload.bb
+++ b/meta-openpli/recipes-connectivity/pyload/pyload.bb
@@ -25,7 +25,7 @@ PR = "r1"
 
 inherit update-rc.d
 
-SRC_URI = "https://github.com/pyload/pyload/releases/download/v${PV}/pyload-src-v${PV}.zip \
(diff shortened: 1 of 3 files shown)


WanWizard:
Why switch pyload to an openelec mirror? 

It is still available at it's source: https://github.com/pyload/pyload/archive/v0.4.9.zip

MastaG:
Yeah I know.. but the filename and checksums differ so I assumed it was a different source.
I'll check it out and make a new PR :)

mthibaut:
I can confirm that this patch works and compiles. But note that libev seems like it will be broken on *every* update upstream. Only the latest libev can be found at http://dist.schmorp.de/libev, but *all* versions of libev can be found at http://dist.schmorp.de/libev/Attic/, so it would seem prudent to me to change the URL to that location so that we don't see this error on every upstream version bump.

PR #230, opened 2017-05-10
lcd4linux: build pydpflib from sources and update to 4.8-r2

athoik:
Requires also https://github.com/eriksl/enigma2-plugin-extensions-lcd4linux-ihad-source-copy/pull/2


eriksl:
Applied.

PR #239, opened 2017-05-17
cdfs: support kernel 4.11

athoik:


littlesat:
Please offer this merge request to develop (so we can cherry pick it to master)

athoik:
There is no cdfs in develop!

WanWizard:
@athoik does that mean develop doesn't support cdfs, or that it is implemented differently?

athoik:
If i remember correctly we are now using gstreamer only!

WanWizard:
Ok, thanks. :+1:

PR #255, opened 2017-06-07
Upgrade exteplayer3 recipe:

MastaG:
- Enable flv2mpeg4 converter
- set package version to match the revision were checking out
- Use fixed srcrev for version 34


meta-openpli/recipes-multimedia/gstreamer/gstreamer1.0_git.bb
```diff
@@ -7,7 +7,9 @@ LIC_FILES_CHKSUM = "file://COPYING;md5=6762ed442b3822387a51c92d928ead0d \
 SRC_URI = " \
 	git://anongit.freedesktop.org/gstreamer/gstreamer;branch=master;name=base \
 	git://anongit.freedesktop.org/gstreamer/common;destsuffix=git/common;name=common \
+	file://deterministic-unwind.patch \
```
athoik: Where did that came from?

MastaG: That's from upstream oe.. see the gstreamer1.0.inc file that got updated.
PACKAGECONFIG[unwind] = "--with-unwind,--without-unwind,libunwind"

However, I wanted to create a sepperate PR using that commit.
I did not know existing PR's would get updated if I make new commits after the initial PR.

So anyways.. yeah. if anybody can apply this.
The upgrade from 1.8 to 1.12 will be fixed :)

athoik: ERROR: gstreamer1.0-1.12.00+gitAUTOINC+7854a65978-r0 do_fetch: Fetcher failure for URL: 'file://deterministic-unwind.patch'. Unable to fetch URL from any source.

PR #257, opened 2017-06-10
Makefile: Add feed switch

persianpros:
It's easier to have a feed switch instead of traditional way.


persianpros:
Why did you close this ?

WanWizard:
I didn't, it probably auto-closed after we had to rebuild the develop branch. And I can't re-open it as the source repo has disappeared.

PR #258, opened 2017-06-10
Keep dm8000 support and delete the rest, Check https://github.com/Ope…

persianpros:
…nPLi/enigma2/issues/920


athoik:
There are people still build images for dm7020hd that has enough storage for develop... Also dm800se can work using flashexpander.

Better to use a BSP layer under OpenPLi and link as submodule.

eriksl:
I do think this is a good idea, but it needs to be discussed within the team.

ims21:
I agree with athoik

persianpros:
@athoik @ims21 You can simply use "meta-local" ;)

athoik:
meta-local for BSP layer?

I prefer meta-dream as submodule instead of deleting some models only.

persianpros:
What I mean is you can what you want with what you have but why keep them on PLi ? ;)

athoik:
ALL OR NOTHING :)

persianpros:
+1 for nothing :D

eriksl:
This was unintentionally closed by WanWizard by remove the relevant branch (which was necessary).

The members have decided that everyone that would like to build unsupported DMM models can always restore the missing models from the BSP. It's no use to us to keep unsupported models forever, and only adds complexity.

So this ís going to be applied, but only if it proves to be a sufficient quality patch.

persianpros:
So what now ? I don't get it because I have to do everything again ...

eriksl:
Applied in rc.

ims21:
Thanks you, very much :-( !

PR #259, opened 2017-06-10
dreambox: restore manufacturer bootlogo

Hains:
Manufacturer bootlogo(first bootlogo) is banned from meta-openpli
https://github.com/OpenPLi/openpli-oe-core/commit/9139f0986a543de8f8d967d09fd33282a7821e3c

Restore it in the place where it belongs.

### DIFF
--- a/meta-dream/conf/machine/dm500hd.conf
+++ b/meta-dream/conf/machine/dm500hd.conf
@@ -17,3 +17,5 @@ CHIPSET = "bcm7405"
 
 # Do not install samba due to small flash size
 IMAGE_INSTALL_remove = "samba-base"
+
+MACHINE_EXTRA_RRECOMMENDS += "dreambox-bootlogo"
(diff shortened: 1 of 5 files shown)


littlesat:
Why do we need this? For every box we do not manipulate manufacturer's boot logo's....

athoik:
Actually we do, if the manufacturer likes to put his splash (in order to overwrite the custom ones), it's his choice and totally "legit". 

Regarding the dreambox, we always used to put manufacturer splash, but that was offered from bootlogo recipe. So that commit simply restores the splash, that used to work under our bootlogo for Dream.

Actually the correct pull request would be "dreambox: restore manufaturer splash".

Hains:
Should i make a new pull-request with proper title ?
I also have some tiny cleanups.

Hains:
Create new one. With new title and some tiny changes.

PR #263, opened 2017-06-14
dreambox: restore manufacturer splash

Hains:
Manufacturer splash is banned from meta-openpli.
See commit:
https://github.com/OpenPLi/openpli-oe-core/commit/9139f0986a543de8f8d967d09fd33282a7821e3c

Wich result in a missing splash after a flash.
Restore splash in the place where it belongs.


eriksl:
Sorry, we decided that DMM should not get a special treatment in this area. We have only Xsarius left that includes their own splash screens and that is the root cause builds for them almost alway fail, so I think they will go at some time as well.

We think a pointer (wiki) to the splash screen plus a step-by-step howto should be sufficient. That also allows one to have it's own splash screen and keep it.

PR #268, opened 2017-06-19
 Allow us to create our own bootlogo's on our stbs

MastaG:
- Enable extra muxers, demuxers, parsers, encoders, decoders and filters
- Build with libfreetype to satisfy the drawtext filter


persianpros:
It won't be accepted because your previous commit is already here : https://github.com/OpenPLi/openpli-oe-core/commit/7a5603a036ba38788dfb428be90ada6e7f0befe8

Just send the https://github.com/OpenPLi/openpli-oe-core/pull/268/commits/7efa60f13633759f5fcf3ff37a708ffa5cd078cd

;)

MastaG:
Yes, I'm still a bit new to git.
However as you can see it will skip the previous commits since they've ready been merged.
There's only one changed file which is from my last commit concerning the ffmpeg bbappend :)

PR #271, opened 2017-06-28
Gstreamer: move the duplicate code into one file

Taapat:
This allows on the update or on corrections perform them in a single file, rather than make the same changes in many files, which can lead to inadvertent errors.


athoik:
Although once we are going to change from 1.12 to 1.14, the branch will be changed in all of them.

Maybe include also GST_BRANCH ?= "1.12" in common?

Taapat:
Good suggestion.

Taapat:
Done, thx!

PR #276, opened 2017-07-01
Keep dm8000 support and delete the rest

persianpros:
More information: https://github.com/OpenPLi/openpli-oe-core/pull/258


ims21:
nonsence... dm800se can boot from usb and still it is good box !
Remove it in your repo!

persianpros:
It's not my opinion ! Check https://github.com/OpenPLi/enigma2/issues/920

persianpros:
Also I did add some changes in comparison to https://github.com/OpenPLi/openpli-oe-core/commit/bf5752b3e805c2c708f46fdecc264bb967010b72

WanWizard:
Some boxes can also boot from NFS. Not really a relevant remark...

The question is whether we as OpenPLi want to keep maintaining EoL hardware when there are more and more specials and hacks needed (like a boot from USB) to keep the image operational. People that know how to create a boot from USB and are smart enough to create their own local build, but why publish an official OpenPLi image with which an enduser can do nothing? And then I'm not even talking about how to support something special like that?

ims21:
Remove dm8000 too ...

PR #284, opened 2017-07-05
meta-dream: remove unnecessary files.

Hains:
Nothing is using it.


WanWizard:
Are you sure? I think there are quite a few DMM plugins that check the presence of tmpd?

Hains:
Third-Party plugins perhaps? totally forgotten.

I will check 3th party plugins and close this PR in that case.

WanWizard:
I don't know, except an old 7000 I've never owned a DMM box. But I vaguely recall this as the reason for leaving it in.

Hains:
Yes i remember that too.
But it should belong to the past if i understand correctly(tpmd = tpm demon).
https://github.com/OpenPLi/openpli-oe-core/commit/061075a5acb0364ebe45cfc7baeeaf8ecb6da38b

But i will check third-party plugins this evening/tomorrow to be sure

Hains:
tmpd is not installed on box. Even not built.
To be sure i  installed and excetuting the plugins:
-dreamexplorer
-dreamirc
- dream concinnity-hd-mod
-dreamexplorer

No problem
BTW: if i try to build tmpd, i get error:
ERROR: Nothing RPROVIDES 'libssl0.9.8x' (but /home/hains/github/openpli-oe-core/meta-dream/recipes-bsp/drivers/tpmd_1.0.bb RDEPENDS on or otherwise requires it)

So it seems harmless to me to remove these recipes.

PR #289, opened 2017-07-07
Samba improvements

Taapat:
See explanations in commits.


WanWizard:
can you revert "samba: add smb.conf in conffiles" from this PR, and send it in on a new PR onto the R/C branch? I think we should fix that for 6.0.

Taapat:
Done!
But I'm afraid that because of the crazy cross dependencies between samba, libsamba and libsamba-base, samba you can not normally update from feed on rc.

Taapat:
This PR should fix these cross dependencies.

eriksl:
All applied to develop.

PR #300, opened 2017-07-16
Update Makefile

persianpros:


athoik:
The develop branch produces the "develop OpenPLi" so calling it 6 doesn't seem right.

WanWizard:
I would say just remove the number, it creates the "OpenPLi environment" no matter the version...

littlesat:
+1 remove the number!

eriksl:
I don't think we need to ditch all DMM's effort. Just leave the copyright in.

Also I don't think we should call it 6 instead of 4. Just leave out the whole release number.

littlesat:
Mmm only the / at the end of dmm's url was taken out...

athoik:
Maybe show branch?

```
@@ -85,13 +85,13 @@ GIT ?= git
 GIT_REMOTE := $(shell $(GIT) remote)
 GIT_USER_NAME := $(shell $(GIT) config user.name)
 GIT_USER_EMAIL := $(shell $(GIT) config user.email)
-
+GIT_BRANCH := $(shell $(GIT) symbolic-ref -q --short HEAD)  hash = $(shell echo $(1) | $(XSUM) | awk '{print $$1}')

 .DEFAULT_GOAL := all
 all: init
        @echo
-       @echo "Openembedded for the OpenPLi 4 environment has been initialized"
+       @echo "Openembedded for the OpenPLi $(GIT_BRANCH) environment has been initialized"
        @echo "properly. Now you can start building your image, by doing either:"
        @echo
        @echo " MACHINE=... make image"
```

athoik:
I add above in https://github.com/OpenPLi/openpli-oe-core/commit/021f9e0905876b273ff47ad9dbb7298c70ba6e0d under test1 branch.

PR #302, opened 2017-07-17
exteplayer3: update to version 35

athoik:


littlesat:
We could consider to merge this... But shouldn't we do in develop instead?

WanWizard:
Depends on whether this is a bugfix version (if so, what does it fix?) or a feature update.

littlesat:
Is this a bugfix version?

athoik:
It a feature, should go on develop.

https://github.com/samsamsam-iptvplayer/exteplayer3/commit/f55def46514c5a93f99bb203ad904dcde6d11e16

PR #307, opened 2017-07-26
Update reporefs.conf

AbuBaniaz:
Make ABM **and** AB 28.2 ${AUTOREV}


WanWizard:
Sorry, no AUTOREV in R/C, only commit hashes. 

The build must remain stable and reproducable, we can't have new updates break an R/C that was working before.

And only bugfixes for the R/C, no new features may be introduced.

AbuBaniaz:
You have other autorev packages!  All the commits in ABM are tested prior to committing.

AbuBaniaz:
By the way, last commit was to workaround a bug in all versions of E2

https://github.com/oe-alliance/AutoBouquetsMaker/commit/89771c6a153f06bb2bd1b7c6b801e58ce6f65e60

I have submitted different pull requests. You might as well stop building ABM if you insist on using older versions and do not want to include bug fixes

Huevos:
How is having and AUTOREV plugin going to make the image unstable?

All commits to this plugin are tested thoroughly.

Plugin depends on us being able to update config files or the data used to create the bouquets becomes stale.

WanWizard:
Because it is OpenPLi who creates the image, and from who users expect the image to be stable. Anyone can say it is tested and stable,  but we are the people users come to complain to. So updating a frozen codebase with new features means the R/C image needs to be fully tested again, on all hardware, not make sure the new feature doesn't break anything. This is extra work we're not prepared to undertake, we don't have the manpower.

So yes, we rather have a tried and tested older version in R/C and Stable, then plugins on the bleeding edge. It is a strategy we've decided upon. Also, don't forget this is the first R/C, and it has taken time to iron all creases out. For next releases I hope there will only be a few days between R/C and Release, and this becomes less of an issue.

The same is true for Release images. There will be no updates for release images, apart from 6.0 to 6.1 to 6.2. So users on a Release will also have to wait for a next release update. That is what the definition of stable is.

Once we have new build resources, we can publish develop images again, and then users will have a choice, either to use stable images, or unstable but bleeding edge images.

Huevos:
Pointless having an out of date version of the plugin. Transponder and service parameters change on a daily basis. Users need to be able to update their configuration files and this is not possible across versions.

I can understand the comments about being stable but the current situation reflects badly on us, the maintainers of the plugin. Pli is forcing users to use an out of date version that contains stale parameters.

WanWizard:
If that is the case a solution needs to be found. The root cause of this issue is then that the data is included with the plugin, instead of kept separate, like is done with the other xml files. So you can't update one without the other.

There will be no intermediate updates of the codebase of a Release image, and we will not deviate from that. We are aiming to have quicker minor releases, it isn't that users will have to wait 12 months after 6.0 before 6.1 arrives.

How do other images this? Are you saying that for example an OpenATV release x.y from today is different from the release x.y from tomorrow? If so, what is the point of using a version number, if x.y != x.y?

WanWizard:
It looks like they compile every night, so a 6.0 from a week ago is not the same as a 6.0 from today. Which means they don't have stable images.

People who want that from OpenPLi, can opt to install a develop image (soon). With all the unstability that comes with that.

Huevos:
I'm from OpenViX. 
Current version is 5.0.
That is followed by build number so current image is 5.0.023.
Then we have intermediate develop builds that are not publicly released, i.e. 5.0.023.005.

Regarding ABM the users can pull the config.xml files from within the plugin but only when the plugin version is current. Version is only bumped when there is a code change which might cause an incompatibility between the .py files and the .xml files, maybe once every couple of months.
The plugin is mature now so most code changes are just minor bug fixes.

Huevos:
So for OpenViX our developer image is built from our E2 Dev branch. When we build a public image Dev is merged to master and we run a build on master. Most of the other gits don't have a develop branch. Core is built from branch 4.0 which is used by most teams and considered stable. And core branch 4.1 is what is being prepared to be the next stable branch. Ultimately the only way to really get code tested is to get it out to the general public.

WanWizard:
Once every couple of months is not a problem, My goal is at least 4 releases per year.

A part of our team has always been against making develop images public, because that would end their "special" status. Since we've become too big to be able to test everything in house, the only solution is to make develop public, as it should be for a true open source project.

PR #309, opened 2017-07-26
Remove ABM

AbuBaniaz:
PLI insists on using outdated version
Avoid downgrades if users installs newer version

### DIFF
--- a/meta-openpli/conf/distro/reporefs.conf
+++ b/meta-openpli/conf/distro/reporefs.conf
@@ -230,7 +230,6 @@ SRCREV_pn-enigma2-plugin-extensions-systemtools = "6008851f1d4556b213fbafaee7340
 SRCREV_pn-enigma2-plugin-extensions-usbformatwizard = "${AUTOREV}"
 SRCREV_pn-enigma2-plugin-extensions-wakeonlan = "ff4efd65559655d4b44991c89dc6506bfb8efb2b"
 SRCREV_pn-enigma2-plugin-extensions-xpower = "3a1e16483589cf4a3a7ae09e8000876a41de66c5"
-SRCREV_pn-enigma2-plugin-systemplugins-autobouquetsmaker = "00251de3c30a14a018cdf19a9efa4f5e6b2efab4"
 SRCREV_pn-enigma2-plugin-systemplugins-autoshutdown = "b0ca70667886d9dd78d7adedca9913ed27a8d2a1"
 SRCREV_pn-enigma2-plugin-systemplugins-crossepg = "${AUTOREV}"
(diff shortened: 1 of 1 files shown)


WanWizard:
Isn't this a very childish response?

AbuBaniaz:
How is it childish? It is pragmatic and user friendly.

By you having an outdated/older version on the plugin server/"feeds", people cannot easily  install the newer version. opkg will not install the newer version because there is an existing/preferred one on the plugin server. People will then have to install using the f-force-overwrite command or start messing around with python files.

You have refused to use the latest STABLE version. 
But you insist users can only install an older version. That is being cruel to users.

It is a plugin people can choose to install, it is not part of the image.

Please remove it and people can manually download ipk from other sources and then install the latest version.

WanWizard:
No, I haven't refused any of that. I have only refused to accept AUTOREV on the R/C branch, and insist on it being pinned on a commit hash.

If there is a latest STABLE version, than that version has a commit hash. So the R/C build can be pinned on that hash without problems. So what is your problem?

AbuBaniaz:
As mentioned in the other rejected commit, you already have other plugins with AUTOREV.

All the python/reader files are tested prior to committing. You want is to commit in the ABM repository, then submit a pull request for PLI to use the newer commit. On every single commit. That is what AUTOREV is designed for. to prevent pointless work.

There has never been an unstable version of ABM in the time I have been testing/contributing to it.

WanWizard:
You don't seem to understand what an R/C is.

The R/C is a test image, pre-release, and under normal circumstances, it will only exist for a week, maybe two weeks, maximum. So there is no need to manually update anything, the image won't live that long.

After the R/C is tested, it will become a Release image, with a fixed version. There will be no daily builds of a release version, it will only be build once. So there as well, no need to update anything.

AbuBaniaz:
I am I am referring to a specific plugin, not the image. 

What is so special about ABM on 7th of May? Any PLI user who wants to install it can only install that one despite the fixes that have gone in afterwards. Whether they use an image built today or on 1st of January 2050, you are dictating that they are pegged to an older version of the plugin.

What is wrong with the fixes and improvements to ABM since 7th of May?

If there is nothing wrong, please use latest one in build process of image. You have already been given a link to the latest  commit. 

If there is something wrong with the fixes, remove the plugin so people can manually install latest one.

AbuBaniaz:
> After the R/C is tested, it will become a Release image, with a fixed version. There will be no daily builds of a release version, it will only be build once. So there as well, no need to update anything.

So all plugins will be checked and last commit hash used before you build the release image?

WanWizard:
Yes.

PR #310, opened 2017-07-27
Added HdmiTest plugin

ims21:
I've placed my plugin (after five year) to git server.

Old ipk can be removed from feeds


WanWizard:
Does that mean it is currently in the 3rd party feed?

Then we have to be a big careful, since at the moment 4 and all 6 images share the same 3rd-partty feed, so it can't be removed yet.

ims21:
Then there in downloadable plugin's list  (Plugins-Green) will be 2 versions ... one without version number (old), one with version number (new). Long time there was AnalogClock too. 
Btw - in Solo4K this plugin missed, as wrote me one user (02.2017) 

Or we can create PR for oe4 too (master?) ? Then could be removed from 3rd feeds

ims21:
Ok, I will close it ...
hdmitest

WanWizard:
No need to close it, we just have to figure out what to do with the 3rd-party feed for OpenPLi 6.

If it is up to me, we empty it, and selectively add things back when people miss them. In which case this PR can be merged without problems.

ims21:
I solve it in my homebuild ... no more needed ...

WanWizard:
So that means for the rest of the world it's ok now? I don't get it?

PR #313, opened 2017-08-13
Fix building OpenPLi with gcc 7

MastaG:
Tested on a clean installation of Fedora 26 using a clean build-directory
- Upgrade libelf from 0.166 to 0.168 (upstream OE)
- Backport upstream OE gcc patch
Tested from a clean state (no build directory) for zgemma h2h (mipsel) and the Mutant HD51 (arm)
Both images build without errors and work perfectly fine.


nickersk:
I am afraid this wont help compiling 3.x kernels ... For example any of Vu+, etc. I think it is caused by terrible devs support from producers. In many cases there seems to be no chance they switch to 4.x kernels in near future...I see the only way to successfull build 3.x is to stick with with gcc6 in machines conf or so...

MastaG:
My apologies I was not really clear.
This PR does not upgrade snapshot of OE nor does it switch to gcc 7.x for building the images.
This is only a fix for two native packages (libelf and the gcc 6.x cross-compiler) which fail to build when the HOST uses gcc 7.

So the compiler built and used by PLi is still gcc 6.

PR #321, opened 2017-08-19
Update enigma2-plugin-extensions-xmltvimport-rytec.bb to 20170819

doglover3920:
pls. include in 60RC and in PLi 4
Thanks


athoik:
Somebody should cherry-pick only https://github.com/OpenPLi/openpli-oe-core/pull/321/commits/05a6f70d370be5ba12778b4701d8759f5c58cbdb

WanWizard:
I'll do that, this is a bit of a mess...

athoik:
In the mean while here is the one #322 for RC-6 (above doesn't apply in RC-6 because it was reverted).

PR #326, opened 2017-08-26
kodi: cleanup platform support

HD-Digital:
Fix FFMPEG build, so we can use external FFmpeg with Kodi and reduce the Kodi patch.


athoik:
Patch seems much cleaner than before, although I have some questions.

1.  For osmega doesn't build, there is an undefined reference to g_v3dPlatformHandle

`make[1]: Leaving directory '/opt/openpli/openpli5test/build/tmp/work/mips32el-oe-linux/kodi/17.1+gitrAUTOINC+7fc6da0c87-r1/git/lib/addons/library.kodi.inputstream'
xbmc/windowing/egl/windowing_egl.a(EGLNativeTypeV3D.o): In function "CEGLNativeTypeV3D::ShowWindow(bool)':
/usr/src/debug/kodi/17.1+gitrAUTOINC+7fc6da0c87-r1/git/xbmc/windowing/egl/EGLNativeTypeV3D.cpp:170: undefined reference to g_v3dPlatformHandle'`

Something needs more care with the patch I guess.

2. Maybe `PACKAGE_ARCH = "${MACHINE}"` should be added in bbappend?

OpenPLi uses shared feeds, that means a package build for mips32el will be the same for all machines. A package build for arm-cortexa again will be shared for all machines.

Above is good if there are no differences between machines. Since we are using different EXTRA_OECONF we need to use MACHINE as PACKAGE_ARCH. Else we are going to get rebuilds if EXTRA_OECONF changes for the same arch.

Of course currently it doesn't conflict, but once another manufacturer uses "another" technique for building Kodi, there will be an issue.

OE-A already uses builds in MACHINE arch due to different patches for the same arch (from different manufacturer).

Thanks.

eriksl:
Agree.

athoik:
Regarding 1. it seems that the following patch is missing.

```
diff --git a/xbmc/platform/xbmc.cpp b/xbmc/platform/xbmc.cpp
index 34e5ea6..42b3d28 100644
--- a/xbmc/platform/xbmc.cpp
+++ b/xbmc/platform/xbmc.cpp
@@ -21,6 +21,9 @@
 #include "Application.h"
 #include "settings/AdvancedSettings.h"

+#include <v3dplatform.h>
+V3D_PlatformHandle g_v3dPlatformHandle;
+
 #ifdef TARGET_RASPBERRY_PI
 #include "linux/RBP.h"
 #endif
@@ -55,6 +58,8 @@ extern "C" int XBMC_Run(bool renderGUI, CFileItemList &playlist)
     return status;
   }

+  V3D_RegisterDisplayPlatform(&g_v3dPlatformHandle, NULL);
+
 #ifdef TARGET_RASPBERRY_PI
   if(!g_RBP.Initialize())
     return false;
```

athoik:
@hd-digital, can you check above notes and provide an update?

athoik:
@HD-Digital, pefroming more tests I notice that `ok` and `exit` buttons do not work on `vs1500` remote control.

Maybe you need to bring back the patches from `xbmc/input/linux/LinuxInputDevices.cpp`?

Can you please check also this one?

HD-Digital:
Edision should merge like other manufactures the v3ddriver header with the EGLNativeTypeV3D Header file.
For now I add a workaround in the patch and succesfull build the osmega image include Kodi.

littlesat:
What about merging this request? Wait for build servers to be stable?

athoik:
We can merge it, it's work in progress, lets see what problems we are going to came up with.

athoik:
@HD-Digital could you please check the kodi for vs1500, hd51?

It seems that kodi cannot start from enigma using kodi plugin (it runs but you get blank screen).

Although if you start kodi (typing kodi on console) while enigma2 runs, it starts fine (you get bootlogo and kodi) but Enigma2 is grabbing the RCU events.

Thanks!

HD-Digital:
@athoik 

The background of the enigma2 kodi plugin should been set on transparent.
I notice that in Openpli there is add two different kodi plugins, one with transparent patch and one without.

https://github.com/OpenPLi/openpli-oe-core/blob/develop/meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-kodi/enigma2-plugin-extensions-kodi_0.1.bb

VS

https://github.com/OpenPLi/openpli-oe-core/blob/develop/meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-kodi.bb

athoik:
@HD-Digital , indeed the transparent patch was the problem getting a black screen. Now it's fine.

Once you play a movie the RCU is not responding any more. Any idea?

Thanks.

PR #329, opened 2017-09-13
[PlexDVRAPI] update to 2.2

rossi2000:


WanWizard:
@rossi2000, for next time, please bump of the PR on a version update. 
@littlesat for next time, please don't accept PR's like this without that bump.

This update has now reset the version to r0.0 (from r0.2), which means opkg will not update to this new release. I'll bump the PR manually this time.

athoik:
@WanWizard, I think you are wrong. Both PV and PKGV updated and that is correct. Since PV/PKGV updated (to higher version) the autopr reset.

WanWizard:
@athoik, Develop builds have failed because of this PR.

> ERROR: enigma2-plugin-systemplugins-plexdvrapi-2.2+gitAUTOINC+6e182e1d58-r0 do_packagedata_setscene: QA Issue: Package version for package enigma2-plugin-systemplugins-plexdvrapi-po went backwards which would break package feeds from (0:2.2+git80+6e182e1-r0.5 to 0:2.2+git80+6e182e1-r0.0) [version-going-backwards]

athoik:
PV changed from 2.1 to 2.2 and is file

auto PR changed from 5 to 0 is the issue, not related with that commit. 

Problem is the DEPENDS = "enigma2". Most probably we really need RDEPEND on enigma2.

WanWizard:
You're much more experienced in these things, but that does sounds very logical. Perhaps the depends can even be removed, none of the other plugins seem to have an `enigma2` dependency.

athoik:
I removed the wrong depend on enigma2, so most probably builds should be fine now.

Once plugin updated again, we should bump PV/PKGV and completely remove PR (since we use auto PR).

In any case, bumping the version is not really required since we are using GITPKGV (as long commits increase by the time) and AUTOREV.

But it's nice that we solved another depend issue!

WanWizard:
I've also noticed the number of MACHINE packages has reduced now, cool!

PR #330, opened 2017-09-26
gstplayer: add shebang and force symlink

devclone:


littlesat:
Why is it required?

devclone:
mistake

PR #331, opened 2017-09-26
plexdvrapi.bb: convert plugin to distutils-openplugins compatible layout

Huevos:


athoik (commented):
Just a minor note, we use SRCREV in reporev file.

I can remove it later.

pbazey:
what it is plz? I do not have a cloud!!!



Sent from Outlook<http://aka.ms/weboutlook>


________________________________

PR #334, opened 2017-10-04
linux-firmware and qthbbtv

zgemma-star:


littlesat:
Is this only for zgemma? Or does it work on more ARM boxes?

PR #335, opened 2017-10-05
Revert "Set /etc/init.d/softcam softlink defaulty to oscam start/stop…

WT-E:
… script"

This reverts commit 4d9d5ec0e2ba9c69b81ff3ef03c4dc6a95b14bbc.


WT-E:
https://forums.openpli.org/topic/53890-bug-softcam-support/

softcam.oscam isn't available in a new installation. It will brake when oscam is not installed, the autoload of other softcams.

PR #347, opened 2017-10-23
Update ffmpeg dash-demux patch and fix gstplayer

MastaG:
Update ffmpeg dash demux patch from SSS:
- improve live playback on youtube
- thanks for fairbird for adjusting it for ffmpeg 3.3
- See: https://forums.openpli.org/topic/41198-serviceapp-gstplayer-and-exteplayer3/?view=findpost&p=786059

Make sure we replace the existing symlink if it already exists. In case the link is already present, opkg will fail.


athoik:
Thanks!

PR #356, opened 2017-11-15
Add nodejs and python-pyexecjs to the feed

MastaG:
It's a requirement for the MediaPortal plugin.
I can confirm it builds for both mips32el and both arm flavours (the one for Vu+ 4k and the one for hd51/vs1500).

See: https://forums.openpli.org/topic/55184-nodejs/


athoik:
Where is python-package-split.inc?

MastaG:
Whoops forgot it.. it's there now.

PR #360, opened 2017-11-27
openpli.conf: add svn to hosttools

blzr:
using host's subversion required on pyro, to fix ...failed with exit code 127, output:/usr/bin/env: 'svn': No such file or directory' error


persianpros:
Yes we did it here too: https://github.com/DMM-PLi/meta-dream/blob/pyro/conf/layer.conf (pyro branch)

PR #367, opened 2017-12-12
Backup Suite plugin update URL

persianpros:
It's just a ready merge request but please wait for Pedro to inform you.
No need to merge this now, I'll update this with a comment later.


WanWizard:
I've merged because it is officially announced, but I see your fork doesn't exist yet? Is it possible to do so, to avoid homebuilds breaking?

persianpros:
It's ok now :)

PR #368, opened 2017-12-13
Backup Suite: license location update

persianpros:


athoik:
Why https://github.com/persianpros/BackupSuite is forked from https://github.com/openmb/BackupSuite instead of original source?

Strange...

persianpros:
I did fork it from Pedro but when he deleted his repo github shows it like this, It's a github thing I guess ...
As you can see my fork is 129 commits ahead which can't be in less than a day and I have Pedro commits.

athoik:
Better to create a new repo and push commits. Search will work also, now in forked repo it doesn't!

persianpros:
Ok I'll do it next week.

PR #391, opened 2017-12-26
meta-dream: fix no image present

Hains:
IMAGE_FSTYPES is ubifs now with previous commit.
Use ubifs consistent.
Apparently IMAGE_CMD_ubifs also needs '_prepend' as addition to create an image.


Hains:
No..

PR #396, opened 2017-12-31
rocko: glibc: fix memory leak

Hains:
Introduced with glibc 2.26 which has -a perthread cache to malloc- enabled by default.
Apply backport patch from glibc 2.27(developement) to prevent a memory with e.g. OScam.
Without this patch, a small RAM memory box like dm8000 will run of of memory within 12 hours and
kernel will kill process.


eriksl:
Applied, thanks Hains!

PR #403, opened 2018-01-29
add machine features for transcodingsetup

zgemma-star:


eriksl:
We already have MACHINE_FEATURE transcoding.

See 2ed677f.

This is superfluous.

eriksl:
Reverted.

zgemma-star:
MACHINE_FEATURE Transcoding installs a proxy for Vuplus
Manufactures which doesn't need the proxy should only install enigma2-plugin-systemplugins-transcodingsetup

eriksl:
We have reasons to also install streamproxy on machines that do not strictly need it. There is no harm in it anyway. Streamproxy will happily convert any incoming request to an "enigma style" transcoding request when suitable, giving all OpenPLi supported receivers a uniform interface. Which is useful when using smartphone apps, that only support one interface or the other.

So I am not entirely happy with the related commit from PieterG yesterday.

pieterg:
The streamproxy needs stb-specific code (stb_feature_t), so it needs to be extended for each new machine.
That's something I'd rather not have, IMHO new devices which follow existing standards should not need modifications in standard components.
Also, the streamproxy is not required on machines using e2 style transcoding.
IMHO optional components should not be installed in the image by default, they should be available in the feed for people who want to use it.
The streamproxy is a hard requirement only on vuplus stbs, so it will continue to be installed in the image for vuplus machines.

pieterg:
If you really wish to have the streamproxy in the image by default, even for machines which do not need it, I could live with that if you modify the streamproxy so it does not need to be extended for each new machine.
(even though I still think optional components should not be installed in the image by default, but that is just my opinion)

eriksl:
Your point would be valid if all manufacturers / models would be using the same interface for transcoding. And clearly that's not the point. For plain streaming there is already the "generic non transcoding machine" that always works (but never really required, and had a bug in that area which I fixed not so long ago).

A new machine like h9, with yet (on details) another interface, can be implemented in streamproxy perfectly, without needing changes in enigma or plugins. And let's face it, only once, twice a year a new transcoding machine is added. Technically it would be better to have another abstraction layer that defines a method of transcoding and assign that models, as there is some duplication now, but the amount doesn't warrant (yet) the amount of work to implement it

 BTW AFAIK the GB models use the same approach to transcoding as VU+, so it's not unique to VU+.

eriksl:
> IMHO new devices which follow **existing standards** should not need modifications in standard components

BTW there is no such thing in transcoding.

pieterg:
Then GB needs the same change in their BSP layer to get the streamproxy installed

PR #405, opened 2018-02-04
Update reporefs.conf

persianpros:
Backup Suite: Try fixing "UnboundLocalError: local variable 'cmd' referenced before assignment"


persianpros:
It fixes the problem: https://forums.openpli.org/topic/21460-backupsuite/?view=findpost&p=839168 :)

PR #409, opened 2018-02-06
Backup Suite: add bzip2 to RDEPENDS

persianpros:
We need "bzip2" for STBs with "tar.bz2" image extension so make sure about it.
I did test this with develop and I'm not sure that we can use "bb.utils.contains" like this for rc 6.1 or not.


WanWizard:
Do you need bzip2, or bunzip2? As that is already in the image?

persianpros:
bzip2, it makes sure of it :)

PR #416, opened 2018-02-13
Update Makefile

persianpros:
Sort BBLAYERS by name

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,33 +43,33 @@ TMPDIR = $(TOPDIR)/tmp
 DEPDIR = $(TOPDIR)/.deps
 
 BBLAYERS ?= \
-	$(CURDIR)/meta-openembedded/meta-oe \
+	$(CURDIR)/meta-amiko \
+	$(CURDIR)/meta-dream \
+	$(CURDIR)/meta-edision \
+	$(CURDIR)/meta-formuler \
+	$(CURDIR)/meta-gfutures \
+	$(CURDIR)/meta-gi \
+	$(CURDIR)/meta-gigablue \
+	$(CURDIR)/meta-local \
+	$(CURDIR)/meta-miraclebox \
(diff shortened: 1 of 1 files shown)


WanWizard:
This might have side-effects if the order in which layers are defined is significant?

(disclaimer: I don't know if that is the case)

persianpros:
I did test it today and compiled two images, It was fine.

Is there any document about layer orders?

WanWizard:
http://www.yoctoproject.org/docs/current/dev-manual/dev-manual.html#understanding-and-creating-layers

**bitbake-layers show-layers** should show them, with all priorities. The problem mainly exists when you have layers with the same priority, containing the same recipes, as the result is unpredicatable. As far as I can see, the order of definition isn't relevant for this.

However, I also find
> It really depends on order of the layers appearing in BBLAYERS variable in toplevel bblayers.conf file, where bitbake will search for .inc files and others where bitbake uses BBPATH since it will search the directories from first to last as specified in BBPATH Therefore if you want a given layer to be considered high priority for the .inc and .conf etc. then consider it adding at the beginning of BBPATH. For bblayers bitbake will use BBFILES_PRIORITY to resolve the recipe contention so the order of directories in BBFILES does not matter.

Which suggest that the "unpredicability" is related to the order in which equal priority layers are defined in BBLAYERS.

WanWizard:
Also:
>  Locating classes and other files is a bit more complicated than recipes - layer priority is not a factor; instead BitBake uses the first matching file in BBPATH, which is manipulated directly by each layer's conf/layer.conf in turn, thus the order of layers in bblayers.conf is a factor - however, each layer.conf is free to either prepend or append to BBPATH (or indeed do crazy stuff with it). Thus the order in BBPATH might not be exactly the order present in bblayers.conf either.

PR #418, opened 2018-02-16
samba: remove unused and not working init script

Taapat:
Init script smaba added in bb file but does not work because of the use of seed options that our busybox does not support.
We use our own smaba.sh int file, so the smaba init file only confuses users.

Also remove unnecessary init.d directory creation which already created in the bb file.

### DIFF
--- a/meta-openpli/recipes-connectivity/samba/samba_4.%.bbappend
+++ b/meta-openpli/recipes-connectivity/samba/samba_4.%.bbappend
@@ -60,10 +60,10 @@ do_install_append() {
 	rm -fR ${D}${bindir}
 	rm -fR ${D}${sysconfdir}/tmpfiles.d
 	rm -fR ${D}${sysconfdir}/sysconfig
+	rm -f ${D}${sysconfdir}/init.d/samba
 	install -d ${D}/var/lib/samba/private
 	install -d ${D}${sysconfdir}/samba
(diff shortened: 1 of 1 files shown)


meta-openpli/recipes-connectivity/samba/samba_4.%.bbappend
```diff
@@ -60,10 +60,10 @@ do_install_append() {
 	rm -fR ${D}${bindir}
 	rm -fR ${D}${sysconfdir}/tmpfiles.d
 	rm -fR ${D}${sysconfdir}/sysconfig
+	rm -f ${D}${sysconfdir}/init.d/samba
 	install -d ${D}/var/lib/samba/private
 	install -d ${D}${sysconfdir}/samba
 	install -m 644 ${WORKDIR}/smb.conf ${D}${sysconfdir}/samba
-	install -d ${D}${sysconfdir}/init.d
```
WanWizard, line 66: Why remove this if that directory is required on the next line?

Taapat: This is bbappend file.
So we also need to check what is done before in bb file.
There, this directory has already been created for the samba init file: https://github.com/openembedded/meta-openembedded/blob/dfbdd28d206a74bf264c2f7ee0f7b3e5af587796/meta-networking/recipes-connectivity/samba/samba_4.6.2.bb#L120
Why try to do this the second time?

PR #422, opened 2018-02-25
Update enigma2-skins.bb

persianpros:
Switch to https://github.com/opendreambox/enigma2-skins

schwerkraft.elitedvb.net sucks all the time so use github instead.


WanWizard:
Do any of these skins still work?

persianpros:
I have no idea I just fixed the URL fetch problem.

PR #425, opened 2018-02-27
use e2 fonts as basic fonts

zgemma-star:


pieterg:
I will cherry-pick some of the commits.

except:
https://github.com/OpenPLi/openpli-oe-core/pull/425/commits/45feaf1c6a72da815b964c54e1cfc27179b652a5 which has a merge conflict
https://github.com/OpenPLi/openpli-oe-core/pull/425/commits/59d594ac1ae4d26695b6f1291606d06341807dcb and https://github.com/OpenPLi/openpli-oe-core/pull/425/commits/3068bcf8e98d87ae9fa1e1e2548dfddf9324726c because I think busybox itself has an ntpd script which we could install, and you should probably depend on some virtual/ntpclient and let the image or distro decide which timekeeping client should be installed.
In fact, if people use e2 together with stalker, they might not want to use an ntp daemon at all.

pieterg:
BTW: please open separate pull-requests in the future, rather than adding several unrelated commits in a single pull-request.
That should make it easier for us to quickly merge requests which clearly are fixes.
Now your fixes are held up because of other -unrelated- commits which need to be discussed.

pieterg:
I'll also leave out https://github.com/OpenPLi/openpli-oe-core/pull/425/commits/bacc2c11534911f52f098cf030cf3df0650bb7d9 which is probably only needed for Rocko.
(meta-vuplus kodiegl (R)PROVIDES kodiegl, so (R)DEPENDing on kodiegl still works in Pyro, but will fail in Rocko)

pieterg:
I've cherry-picked your fixes, I'll close this request.
please open new, separate, requests for the remaining stuff.

PR #426, opened 2018-03-02
openpli-common.conf: use default FULL_OPTIMIZATION -O2 OE-core

Hains:
For better performance.
This increasing the image size a bit(2.5MB for dm8000).

Small cosmetic change in comment.

Thanks to Athoik.

### DIFF
--- a/meta-openpli/conf/distro/openpli-common.conf
+++ b/meta-openpli/conf/distro/openpli-common.conf
@@ -39,16 +39,8 @@ LICENSE_FLAGS_WHITELIST = "commercial"
 COMMERCIAL_AUDIO_PLUGINS ?= "gst-plugins-ugly-mad gst-plugins-ugly-mpegaudioparse"
 COMMERCIAL_VIDEO_PLUGINS ?= "gst-plugins-ugly-mpeg2dec gst-plugins-ugly-mpegstream gst-plugins-bad-mpegvideoparse"
 
-# OE optimization defaults to -O2 which makes for much larger binaries.
-# Override here to use -Os instead, resulting in smaller images.
-FULL_OPTIMIZATION = "-Os -pipe ${DEBUG_FLAGS}"
-# build some core libs with better compiler optimization for better performance
-O2_OPT = "-O2 -pipe ${DEBUG_FLAGS}"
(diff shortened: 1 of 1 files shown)


athoik:
What about performance? I am sure there will be small gains everywhere (faster python, e2, etc).

eriksl:
I am worried about first gen receivers, like xtrend 9x00 and vuduo, that may not have enough flash.

Hains:
I created an image for et9x00 with the latest development with -O2 as default (note: I had to update e2fsprogs to 1.43.8 and qemu to 2.11.1) and extracted the zip file.

The created et9x00 folder has a size of 95 MB.

About performance:
I did not notice performance improvement with dm8000.
O2 will probably be faster, but we're not talking about seconds but about milliseconds.

With -O3 as default I notice a little less delay in the use of gstreamer.
Somewhat faster startup of streams / files.
And for example, a restart of enigma2:
With -Os and -O2 as default: +/- 42 seconds.
With -O3 as default: +/- 35 seconds.

So, -O3 has a speed improvement of about 16%.
But from O2 to O3 as default , increases the image size by 3MB.

eriksl:
How much increase in image size from Os to O2?

RobvanderDoes:
> I am worried about first gen receivers, like xtrend 9x00 and vuduo, that may not have enough flash.

After confirming that this is indeed an issue: why not stop building those images? You can't support ancient boxes forever; there comes an end to it, otherwise innovation can never take place.

Hains:
```
How much increase in image size from Os to O2?
```

With -Os: (without this PR)   91.9MB   (zipfile 70.9MB).
With -O2: (with this PR)        95.0MB   (zipfile 73.3MB)

So 95-91.9 = 3.1MB.

eriksl:
That might make the difference voor older generations receivers. We may consider not apply it for them.

littlesat:
+1,
Unless find a way to implement it only for the boxes with bigger flash...

eriksl:
There is a way, but I don't like it:
- it requires per-machine overrides, these belong in the BSP
- it may result in bugs showing in some machines and not in others, difficult to debug

athoik:
Maybe switch to O2 as default and change E2 to O3?

PR #436, opened 2018-03-12
libdvbsi: Switch to PLi-metas as git.opendreambox.org is not reachable, Also all patches applied.

persianpros:


athoik:
It seems now it's up.

PR #439, opened 2018-03-16
[.gitmodules] Use https for Axas-UHD

AbuBaniaz:
This avoids people having to have Git credentials when building. All others also use https


AbuBaniaz:
Closed because of this
https://github.com/OpenPLi/openpli-oe-core/commit/e9969d09a2e23948ee129671cc80b188f37058c4

PR #455, opened 2018-04-04
oscam: Switch to metas-openpli

Hains:
Temporary solution for the time being, until
OScam is back online with its own repo.

-Update OSCam to 11402.
-Update OSCam-emu to r769.

### DIFF
--- a/meta-openpli/conf/distro/reporefs.conf
+++ b/meta-openpli/conf/distro/reporefs.conf
@@ -211,8 +211,8 @@ SRCREV_pn-enigma2-plugin-picons-ziggo.metal-stamp-100x60 = "${AUTOREV}"
 SRCREV_pn-rtl8188eu = "${AUTOREV}"
 SRCREV_pn-rtl8192eu = "${AUTOREV}"
 
-SRCREV_pn-enigma2-plugin-softcams-oscam ?= "${AUTOREV}"
-SRCREV_pn-enigma2-plugin-softcams-oscam-emu ?= "11400"
+SRCREV_pn-enigma2-plugin-softcams-oscam = "5e695a59383af46fce24309040f40bfbb6dde127"
+SRCREV_pn-enigma2-plugin-softcams-oscam-emu = "0553e3d382726955379046a0838b7110554ded2e"
 SRCREV_pn-blindscan-s2 = "${AUTOREV}"
 
 # e2openplugins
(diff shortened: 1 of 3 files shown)


meta-openpli/conf/distro/reporefs.conf
```diff
@@ -211,8 +211,8 @@ SRCREV_pn-enigma2-plugin-picons-ziggo.metal-stamp-100x60 = "${AUTOREV}"
 SRCREV_pn-rtl8188eu = "${AUTOREV}"
 SRCREV_pn-rtl8192eu = "${AUTOREV}"
 
-SRCREV_pn-enigma2-plugin-softcams-oscam ?= "${AUTOREV}"
-SRCREV_pn-enigma2-plugin-softcams-oscam-emu ?= "11400"
+SRCREV_pn-enigma2-plugin-softcams-oscam = "5e695a59383af46fce24309040f40bfbb6dde127"
```
persianpros, line 214: What's wrong with AUTOREV when it's on our git?
Now we're on svn11409: https://github.com/PLi-metas/oscam

Hains: To match with svn revision..If we use AUTOREV we have to drop svn revision.
So e.g. oscam-1.20-git130+afda92b-r0
Or we need to find a way to grab 'svn revision' from GIT if possible?

MastaG:
Unless we're going to fork oscam as well, this sounds like a proper solution.
We from PLi-metas are just mirroring oscam for the moment, we won't do any unofficial changes to it.

meta-openpli/conf/distro/reporefs.conf
persianpros: Can you grab it from releases page: https://github.com/PLi-metas/oscam/releases ? But I have to create them each time.

Also after each commit there is this line: git-svn-id: http://www.streamboard.tv/svn/oscam/trunk@11410

I mean "trunk@number"

persianpros:
Yes it's only a mirror nothing more and if we ever going to do anything we'll create a fork and rename it :)

meta-openpli/conf/distro/reporefs.conf
WanWizard: Are you using the same mirror tools as oscam.cc used to use? If so, I have added some updates (commited by theparasol) to get the SVN number from the git-svn-id in the commit message (see https://github.com/PLi-metas/oscam/commit/d3b3898cee0b1f4065c689b9b9a4c16d034ce660) so AUTOREV should work fine).

PR #456, opened 2018-04-04
exteplayer3: update to git HEAD master

Hains:
No exteplayer has patches for ffmpeg 3.4.2
https://github.com/samsamsam-iptvplayer/exteplayer3/commit/b2e73a598378ceab3ea15b043e7b66c7fe14c3c0

Added linuxdvb_buffering.c to source, see commit:
https://github.com/samsamsam-iptvplayer/exteplayer3/commit/86c7be1cdea3a05c9b3edcd845c3d24c9a2d4499

No tested for sh4.

### DIFF
--- a/meta-openpli/conf/distro/reporefs.conf
+++ b/meta-openpli/conf/distro/reporefs.conf
@@ -38,7 +38,7 @@ SRCREV_pn-enigma2-plugin-settings-hans-kabelnl = "${AUTOREV}"
 SRCREV_pn-enigma2-plugins = "${AUTOREV}"
 
 SRCREV_pn-libdreamdvd = "${AUTOREV}"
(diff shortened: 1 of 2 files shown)


MastaG:
Tested on all platforms :)
I can confirm the output/linuxdvb_buffering.c source file is only meant for broadcom soc's (as it doesn't compile on sh4).

Please push this asap as we're already shipping ffmpeg 3.4.2 on develop.

One thing we could do perhaps is just remove the version number from the recipe.
samsamsam bumps it for every commit he does.
It's now at version 44.
While we still have:
PV = "38+gitr${SRCPV}"
PKGV = "38+gitr${GITPKGV}"

So if we're going to AUTOREV which is fine by me, then we might as well remove this number.

PR #462, opened 2018-04-19
Add optional systemplugin to install support for BH skins

MastaG:
These files used to be in OpenBlackHole's public CVS, which was forked from PLi's enigma2.


littlesat:
Should we really add this this way?

littlesat:
Bh skin are we allowed to add then in the first place?

MastaG:
Hi, these files were part of the OpenBlackHole enigma2 CVS, this was forked from our PLi a long time ago.
So these files are GPLv2+.

It will actually install the exact same files as here: https://github.com/OpenPLi/enigma2/commit/4a44f93353dbac2c0c917cd5b6fedc403389bbde

With the only difference being their original file-names and class-names, to make BH skins work out of the box.

So to sum things up:
1. We ARE allowed to ship these files.
2. We already HAVE these two files in our enigma2 core, but with generic file- and class-names.
3. This is only an optional compat-package (in the feed only) for people who would like to use BH skins.

PR #472, opened 2018-05-21
Update reporefs.conf

PD1LOI:


PD1LOI:
Small update in RDS display

PR #480, opened 2018-06-02
zgemma: update bsp

zgemma-star:


eriksl:
Why?

What does it do?

Why not a normal BSP submodule update?

zgemma-star:
Yes, this was a request of submodule update for openpli 6.2
But new update are available already so this is out of date.

zgemma-star:
With update branches for openpli 6.2 it will probably fix the qt build and add support for all available zgemma models.

WanWizard:
We have an automated update script that updates the submodules, which we run at least once a week. I have just started in manually so the BSP submodules in develop are up to date again.

zgemma-star:
Yes, develop branch is uptodate but latest image Openpli6.2 isn't update anymore for the new models and fixes

WanWizard:
Correct, release images are marked "stable", only bugfixes are applied. Updating a BSP might have impact elsewhere (think about webkit or hbbtv issues) which means we need to run full-blown tests again to make sure there is no end-user impact.

zgemma-star:
When the BSP layer is not update anymore for the release image when will come Openpli 6.3?
Now most release images use drivers from 6 months or older.

WanWizard:
As soon as Sumo is ready to become the next develop, develop will become the next release. The version hasn't been decided yet, as 6.x currently uses Morty, but develop is on Pyro. We're working on it.

PR #483, opened 2018-06-12
Update openpli-common.conf

persianpros:
I know PLi won't support dm500hd/dm800/dm800se anymore but with this configuration we can build compressed images again, see: https://github.com/PLi-metas/pli-extras#how-to-build-compressed-images-for-smallflash-stbs

It won't harm anything and it's a good thing to have what could work in this file as a reference for other developers.

### DIFF
--- a/meta-openpli/conf/distro/openpli-common.conf
+++ b/meta-openpli/conf/distro/openpli-common.conf
@@ -47,6 +47,9 @@ COMMERCIAL_VIDEO_PLUGINS ?= "gst-plugins-ugly-mpeg2dec gst-plugins-ugly-mpegstre
 # UPDATE: We don't support receivers with small flash anymore.
 # So default to 02 for faster code
 FULL_OPTIMIZATION = "-O2 -pipe ${DEBUG_FLAGS}"
+FULL_OPTIMIZATION_dm500hd = "-Os -pipe ${DEBUG_FLAGS}"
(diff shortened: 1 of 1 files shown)


ims21:
I think, you was first, who removed this old boxes, if I see dmm oe-core history...

persianpros:
@littlesat yes it works for all small flash boxes as we do this for azboxhd too.

You can build and use Open PLi 6.x with pli-extras and see for yourself but we use UPX compression too ;) So FULL_OPTIMIZATION and UPX.

UPX example: https://github.com/PLi-metas/pli-extras/blob/develop/recipes-multimedia/exteplayer3/exteplayer3_git.bbappend

What UPX does: https://github.com/PLi-metas/pli-extras/blob/develop/classes/upx_compress.bbclass

@ims21 I just helped PLi with the cleanup, I didn't decide it, you can build PLi for all dreamboxes even arm models using our git: https://github.com/PLi-metas/pli-extras

We help people to have PLi experience but the DMM company won't.

pieterg:
There should be no machine-specific stuff in meta-openpli. Stuff like that belongs in a bsp.

ims21:
@PP ... i am using in my git build for my dm800se (  I am using dm800se as old storage with recorded movies only), but I am not using flash, I put all to usb stick.

PR #485, opened 2018-06-13
Update openpli-common.conf

persianpros:
Add 'LOCALE_GENERATION_WITH_CROSS-LOCALEDEF_forcevariable = "0"' for recent linux distributions.

### DIFF
--- a/meta-openpli/conf/distro/openpli-common.conf
+++ b/meta-openpli/conf/distro/openpli-common.conf
@@ -99,3 +99,7 @@ PACKAGECONFIG_pn-neon = "expat openssl libproxy webdav zlib"
 
 # Allow java, svn and unzip executed on host
 HOSTTOOLS += "java svn unzip"
+
+# If you're going to build on a linux distro which features glibc 2.26 or newer (Ubuntu 18.04 LTS for example),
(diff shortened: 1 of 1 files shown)


persianpros:
See: https://github.com/PLi-metas/pli-extras/blob/develop/conf/layer.conf#L9

eriksl:
What does it do.

persianpros:
If you're going to build on a linux distro which features glibc 2.26 or newer (Ubuntu 18.04 LTS for example), then uncomment below line, more information: https://bugzilla.yoctoproject.org/show_bug.cgi?id=12265
# LOCALE_GENERATION_WITH_CROSS-LOCALEDEF_forcevariable = "0"

betacentauri:
And why don’t you add the real fix mentioned in the bugzilla ticket?

persianpros:
Ok I just did for our git: https://github.com/PLi-metas/pli-extras/commit/3e9cc8da86f62b986d4459851d66dd712281da3b

After test I'll let you know :)

persianpros:
@betacentauri 
It didn't work, the patch applied but gives compile error so I did revert it.

PR #489, opened 2018-06-13
Update layer.conf

persianpros:
I know PLi won't support dm500hd/dm800/dm800se anymore but with this configuration we can build compressed images again, see: https://github.com/PLi-metas/pli-extras/tree/8ea9b0e7cc32c693af6106210c4712dc2201860c#how-to-build-compressed-images-for-smallflash-stbs

It won't harm anything and it's a good thing to have what could work in this file as a reference for other developers.

It works for all small flash boxes as we do this for azboxhd too.

You can build and use Open PLi 6.x with pli-extras and see for yourself but we use UPX compression too ;) So FULL_OPTIMIZATION and UPX.

UPX example: https://github.com/PLi-metas/pli-extras/blob/develop/recipes-multimedia/exteplayer3/exteplayer3_git.bbappend

What UPX does: https://github.com/PLi-metas/pli-extras/blob/develop/classes/upx_compress.bbclass

### DIFF
--- a/meta-dream/conf/layer.conf
+++ b/meta-dream/conf/layer.conf
@@ -2,3 +2,13 @@
 BBPATH .= ":${LAYERDIR}"
 
 BBFILES += "${LAYERDIR}/recipes-*/*/*.bb"
+
+FULL_OPTIMIZATION_dm500hd = "-Os -pipe ${DEBUG_FLAGS}"
+O2_OPT_dm500hd = "-Os -pipe ${DEBUG_FLAGS}"
+O3_OPT_dm500hd = "-Os -pipe ${DEBUG_FLAGS}"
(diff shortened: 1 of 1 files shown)


persianpros:
See: https://github.com/PLi-metas/meta-dream/blob/develop/conf/layer.conf#L14~L22

eriksl:
I don't think layers.conf is the appropriate place for this. Home builders can add this to their local.conf.

betacentauri:
See also this comment from MiLo:
That's silly. The flash filesystem is already compressed.

Besides, compressed binaries use physical RAM, since they cannot be swapped and shared. Libraries and other executables are "mmapped" in shared mode, so there's only one copy in RAM. If you compress it, each process gets its own private (decompressed) copy. Bad. Don't do it.

If you go the "compression" way, you'd better go for XZ compression on ubifs, I think there's a patch to port that to most kernels

PR #499, opened 2018-07-01
Add libvpx 1.6.1 as the old version won't compile in Ubuntu 18.04 LTS

persianpros:


eriksl:
Shouldn't it replace the old version then?

persianpros:
I don't know, we use it like this in pli-extras: https://github.com/PLi-metas/pli-extras/tree/develop/recipes-multimedia/webm

eriksl:
That's not OpenPLi's concern ;-)

persianpros:
Yeah but without this we can't compile on Ubuntu 18.04 LTS.

eriksl:
AFAIK OpenPLi doesn't use vpx at all?

If that's the case, I suggest replacing it instead of adding it.

PR #502, opened 2018-07-02
Update softcams-enigma2-meta.bb

persianpros:
We don't have an "aarch64" binary for CCcam.

### DIFF
--- a/meta-openpli/recipes-openpli/enigma2-softcams/softcams-enigma2-meta.bb
+++ b/meta-openpli/recipes-openpli/enigma2-softcams/softcams-enigma2-meta.bb
@@ -14,7 +14,7 @@ DEPENDS_append_mipsel = "\
 
 # other softcams
 DEPENDS += " \
-	enigma2-plugin-softcams-cccam \
+	${@bb.utils.contains("TARGET_ARCH", "aarch64", "", "enigma2-plugin-softcams-cccam", d)} \
 	enigma2-plugin-softcams-oscam \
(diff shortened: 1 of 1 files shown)


eriksl:
???

persianpros:
It's for this problem: https://forums.openpli.org/topic/60847-failed-to-build-feeds-for-aarch64-and-clean-the-build-directory/?view=findpost&p=904249

eriksl:
We don't support 64 bit in any way yet, and maybe we won't do it at all. See 64 vs 32 bit on STB discussion.

persianpros:
Yeah I know but it's on develop branch for extra development like what we do on pli-extras, anyway it's your choice to accept this :)

eriksl:
Can't you do this in the bsp?

moham96:
>  See 64 vs 32 bit on STB discussion.

@eriksl  can you link me to the discussion?

eriksl:
In a nut shell: on x86 it helps to use 64 bit. Not because of the 64 bit, but because AMD64 also extends the register set and adds some smart instructions. Your code will run faster on 64 bit.

On ARM there is no benefit in using the 64 bit instruction set. The code will not run faster, only slower because all pointer read/writes to/from memory will need to access 8 instead of 4 bytes. Also the code will be larger. 64 bit will only be interesting if you want > 32 bit virtual address space, i.o.w. more than 4 Gb per process. That's not quite likely on a STB.

PR #505, opened 2018-07-18
Create cleanup-build.sh

persianpros:
Sometime we compile so many images for so many STBs but we don't want to keep them all, also we don't want to remove the build folder too as we have to compile everything again in that case!
With this script you just give the machine name and it will cleanup the build folder for you faster than you could imagine,
so the size of the build folder will be reduced and you will have more space again.


persianpros:
We use this: https://github.com/PLi-metas/pli-extras/blob/develop/cleanup-build.sh

PR #509, opened 2018-08-05
dvbsnoop.bb: Version 1.4.54

persianpros:
New: dvb_api.h is compatible with old V4L DVB headers now.
New: Experimental BBFrame filtering support. (Reverted)
Fix: Add missing autogen.sh


persianpros:
@WanWizard It's just a version change :)

PR #513, opened 2018-08-07
Serviceapp requires curl for script support

MastaG:
ServiceApp recently got support for running scripts.
So we also install curl, as this is mostly used for url processing.


athoik:
User script can have lot of depend's.

So user should handle depends, or script should be part of OE build, and there proper depends can be used.

PR #519, opened 2018-08-28
Update reporefs.conf

PD1LOI:
Aanpassing voor Filebrowser


PD1LOI:
Update Filebrowser in the skin.

PR #522, opened 2018-08-30
Inadyn mt

pr2git:
Inadyn-mt no longer compile because the version no longer exist.
So I bring back here the OE-Alliance version which compile fine.
All credits goes to OE-A.


meta-openpli/recipes-connectivity/inadyn-mt/inadyn-mt.bb
```diff
@@ -2,17 +2,17 @@ MAINTAINER = "Narcis Ilisei"
 LICENSE = "GPLv3"
 LIC_FILES_CHKSUM = "file://COPYING;md5=3c34afdc3adf82d2448f12715a255122"
 
-PV = "v.02.24.44"
-PR = "r3"
+PV = "v.02.28.10"
+PR = ""
```
WanWizard: Should this be empty, instead of "r0"?

pr2git: Please revert it, I think I create a wrong PR.

PR #558, opened 2018-10-03
gstreamer: update to version 1.14.4

Hains:
Highlighted bugfixes in 1.14.4

-glviewconvert: wait and set the gl sync meta on buffers
-glviewconvert: Copy composition meta from the primary buffer to both outputs
-glcolorconvert: Don't copy overlay composition meta over to NULL outbufs
-matroskademux: add functionality needed for MSE use case fixing youtube playback in epiphany/webkit-gtk
-msdk: fix build on windows
-opusenc: fix another crash on 32-bit x86 on windows (alignment issue in SSE optimisations)
-osxaudio: add support for parsing more channel layouts
-tagdemux: Use upstream GST_EVENT_STREAM_START (and stream-id) if present
-vorbisdec: fix header handling regression: init decoder immediately once we have headers
-wasapisink: recover from low buffer levels in shared mode
-fix GstSegment unit test which would fail on some 32-bit x86 CPUs


Hains:
Sorry,  mistake SRCREV_ffmpeg_pn-gstreamer1.0-libav

PR #561, opened 2018-10-04
pyton-twisted: replace underscore with dash

Hains:
To fix build error:

ERROR: Nothing RPROVIDES 'python-service_identity' (but /home/hains/openpli-oe-core/
meta-openpli/recipes-devtools/python/python-twisted_17.9.0.bb RDEPENDS on or otherwise requires it)


WanWizard:
Thanks. Never commit when you're in the pub... ;-)

PR #564, opened 2018-10-20
init-ifupdown_1.0.bbappend: Fix parse error

persianpros:


ims21:
SRC_URI +=

persianpros:
Yeah I have a reason for "append" and "+", you can check it yourself as I did and it's ok ;)

eriksl:
You're not supposed to use += on an _append.

WanWizard:
Never commit things late at night..., should indeed be +=.

PR #577, opened 2018-11-06
It's not necessary

persianpros:
It's not necessary


betacentauri:
We don't think that is necessary. See also comments here:
https://github.com/oe-alliance/oe-alliance-core/pull/414

PR #584, opened 2018-11-20
Fix mdev.conf to load duo4k wifi firmware and driver.

vu-plus:
Fix mdev.conf to load duo4k's wifi firmware and drivers.

Unlike udev, mdev uses a single config file (/etc/mdev.conf), so the duo4k wifi problem could not be resolved in meta-vuplus BSP Layer.


WanWizard:
Thanks!

Added to the rc branch, and merged back into develop.

PR #585, opened 2018-11-21
Fix mdev.conf for ultimo4k/duo4k.

vu-plus:
Revert previous commit 896c140cc9d696ac9e2afba28a41f798ef38f608.
And add fixed commit, Fix to mdev.conf do not apply ultimo4k, but only duo4k.

### DIFF
--- a/meta-openpli/recipes-core/busybox/busybox/mdev.conf
+++ b/meta-openpli/recipes-core/busybox/busybox/mdev.conf
@@ -49,7 +49,4 @@ dboxlcd         0:0     660     =dbox/lcd0
 pvr             0:0     660     =misc/pvr
 vtuner([0-9])   0:0     660     =misc/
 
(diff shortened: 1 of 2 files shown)


WanWizard:
This does not address the problem, which is:
> log.do_rootfs.29124:update-rc.d: ~/develop/build/tmp/work/vuultimo4k-oe-linux-gnueabi/openpli-enigma2-image/1.0-r0/rootfs/etc/init.d/vuplus-wifi-init.sh: file does not exist

which is the init script to enable wifi on the ultimo4k ?

As far as I can see this means it is no longer present in the archive that is downloaded from your server in the recipe?

vu-plus:
Sorry, I found wrong part in meta-vuplus. 

New commit is updated in meta-vuplus. Please apply it.

Thank you.

WanWizard:
Thanks for the quick response. 

Implemented, and a new 7.0-rc put online. Awaiting tester feedback.

PR #586, opened 2018-11-21
Fix exFAT filesystem support

Taapat:
For exFAT file system mount is used fuse-exfat module.
Without it, the exFAT mount does not work.
After the fuse-exfat installation, also works the standard 'mount -t auto'.

Therefore, add fuse-exfat in openpli-image and remove the unnecessary exception for exFAT mount in mdev-mount.sh.
Also, the correct mount.exfat path is /sbin/mount.exfat instead of /usr/sbin/mount.exfat, so the exception in mdev-mount.sh does not work in any case.

### DIFF
--- a/meta-openpli/recipes-core/busybox/busybox/mdev-mount.sh
+++ b/meta-openpli/recipes-core/busybox/busybox/mdev-mount.sh
@@ -155,9 +155,6 @@ case "$ACTION" in
 				MOUNTPOINT="/media/$MDEV"
 				mkdir -p "${MOUNTPOINT}"
 			fi
-			if [ $TYPE == exfat ] && [ -f /usr/sbin/mount.exfat ] && mount.exfat -o noatime /dev/$MDEV "${MOUNTPOINT}" ; then
-				exit 0
-			fi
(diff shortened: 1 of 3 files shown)


betacentauri:
Is fuse-exfat always available?

WanWizard:
Look like it:
````
./mips32el/fuse-exfat_1.2.3-r0.0_mips32el.ipk
./cortexa15hf-neon-vfpv4/fuse-exfat_1.2.3-r0.0_cortexa15hf-neon-vfpv4.ipk
./armv7ahf-neon/fuse-exfat_1.2.3-r0.0_armv7ahf-neon.ipk
./mips32el-nf/fuse-exfat_1.2.3-r0.0_mips32el-nf.ipk
./aarch64/fuse-exfat_1.2.3-r0.0_aarch64.ipk
````

Taapat:
This is not a driver from kernel but aditional tool: https://github.com/openembedded/meta-openembedded/blob/pyro/meta-filesystems/recipes-filesystems/fuse-exfat/fuse-exfat_1.2.3.bb

PR #593, opened 2018-12-01
create share conf files for devices mounted in fstab too

ims21:


littlesat:
Why?

ims21:
Why not ?

littlesat:
As I do not think they should be added...

ims21:
I will not pray for it ...

littlesat:
Then firsr explain what is the problem/need and/or what solved is?

WanWizard:
If this is for local devices only, I can live with it, but I am against it for remote devices. It it pointless to have a box re-share all my NAS mounts for example...

ims21:
Your mounted remote devices are in /proc/mounts  as /dev/$MDEV ?

WanWizard:
Eh... I sure hope not. You've got a point.

I do have an issue with line 176: that is an else executed when the mount of the added device fails. How can it ever get there for fstab defined mounts?

ims21:
I added it, when I found, that on some boxes (f.eg. my vs1500) is not for UUID=xxxxxx /media/usb in fstab created usb.conf, because line 123 is false - due race ?

WanWizard:
Can you give the exact line that fails for you? 
I'll try to check if it's a race or something else. If not a race, it should be fixed and not worked around.

p.s. why would you want to mount a USB device on UUID? Why not use the preferred way, a label?

ims21:
Why not ? When I put next USB stick and reboot box, stick with UUID will be mounted always same.  
For line:
UUID=4CD6-2774 /media/usb auto defaults 0 0
is not created .conf, because line 123 is false, because /dev/sdb1 (my usb with UUID) is founded as line in fstab - weird, that is not found as mounted in line 64 (same as hdd) in /proc/mounts
But when box is started, /dev/stb1 in /proc/media exist

PR #602, opened 2018-12-17
autobouquets: update package version 2.2

LraiZer:


WanWizard:
No need to create a PR for develop if there was one for RC. RC will be merged back at regular intervals.

PR #611, opened 2018-12-25
Update enigma2-plugin-extensions-backupsuite.bb

persianpros:
We don't provide support on sumo anymore.
You could ask https://github.com/Hains for that specific commit which makes it compatible with sumo.


WanWizard:
What does this mean? There will not be a working backupsuite anymore when we switch to sumo?

PR #614, opened 2018-12-29
Change PLi-metas to OpenVisionE2

persianpros:
Sorry for any inconvenience but it has to be done as we can't use PLi names.
This is urgent.
@WanWizard @littlesat @eriksl


WanWizard:
Manually applied to the RC branch, and merged back into develop.

PR #622, opened 2019-01-11
Plugin youtube: fix live stream extraction

Taapat:
Fix threading in GoogleSuggestionsConfigText


WanWizard:
Is this also needed in rc ?

Taapat:
This is for RC ;)

U.P.D.
Sorry I add in rc too.

PR #624, opened 2019-01-11
Update lsdir_git.bb

kueken:
switch Url &Thanks for Recover @captain (ATV)


WanWizard:
No thanks, not really interested in using a fork of a fork. 

If it stays offline, we'll see if we setup our own copy, or (and that is my preference) remove it from the build, it is a bit pointless if the author/maintainer has removed himself from the internet...

/cc @eriksl

kueken:
constructively one can not call that probably...
I do not need to update the other things also.
This would only be temporary, since the git of samsamsam works.
But OK.
you do not need help.

WanWizard:
Bit of an odd reaction. Long toes?

PR #625, opened 2019-01-12
Update url e2iplayer dependencies

Hains:
samsamsam-iptvplayer => e2iplayer.

Repo samsamsam-iptvplayer is deleted.

### DIFF
--- a/meta-openpli/recipes-multimedia/exteplayer3/exteplayer3_git.bb
+++ b/meta-openpli/recipes-multimedia/exteplayer3/exteplayer3_git.bb
@@ -11,7 +11,7 @@ inherit gitpkgv
 PV = "50+gitr${SRCPV}"
 PKGV = "50+gitr${GITPKGV}"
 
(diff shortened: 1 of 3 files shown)


WanWizard:
We've noticed, very handy. Not. :-1: 

Thanks!

Hains:
But no lsdir. 
We have (temporarily?) 
https://github.com/OpenVisionE2/lsdir

Is that acceptable?

WanWizard:
Just fixed.

It is only a single c source file, which is old (not from samsamsam anyway), and will never change, so we've opted for a tarball instead of a repo dependency.

WanWizard:
Appearently two more are about to go (but fortunately not gone yet):
> gstplayer: gone to https://gitlab.com/e2i/gstplayer
>gst-ifdsrc: gone to https://gitlab.com/e2i/gst-ifdsrc

would you be able to have a look at those? This seems to be more than simply changing repo's, with my very limited knowledge, I can't get them to compile...

PR #628, opened 2019-01-15
gst-ifdsrc: update url

Hains:
samsamsam => e2i


WanWizard:
Can't believe it was that simple, I couldn't get it to compile ! Thanks again !

PR #644, opened 2019-02-14

doglover3920:
World of satellite new URL


doglover3920:
Do not merged.  Made an error

PR #646, opened 2019-02-14

doglover3920:
World of sat URL changed


WanWizard:
ERROR: enigma2-plugin-extensions-epgimport-rytec-20190214-r0 do_fetch: Fetcher failure for URL: 'http://www.vuplus-community.net/rytec/sources/rytec.sources.xml.20190214.gz'. Unable to fetch URL from any source.

PR #647, opened 2019-02-14

doglover3920:
World of Sat URL changed


WanWizard:
**NOTE:** Don't merge this, the build will fail!

PR #651, opened 2019-02-19
built all the necessary qt packages as machine specific

zgemma-star:


WanWizard:
Sounds like a plan, thanks!

PR #653, opened 2019-02-19
Sumo eudev revert

zgemma-star:
eudev seems not ready yet and often is clears all the data of the usb storage devices.
I would suggest to use mdev for the time being.

it would been nice if sumo branch will turn to develop branch and that we can fix the final built issues.


WanWizard:
I think this is being looked at. @eriksl @betacentauri ?

betacentauri:
We are aware of this. First analysis was done. Populate-volatile.sh deletes the data on first boot after flashing when it creates /run/media link.

zgemma-star:
So why not make a subbranch for eudev and systemd development till it become functional. 
Mdev always did the job well and in sumo it still does.
It seems to finalize sumo branch with mdev is not a huge job anymore as most is done already.

littlesat:
As far I understand we also wanted to change to udef with sumo+

PR #677, opened 2019-04-19
Use shallow clone for git modules

persianpros:
There's no need for the history of companies so fetch them faster just like https://github.com/OpenVisionE2/openvision-oe/blob/develop/.gitmodules


WanWizard:
This requires a pretty recent version of git. CentOS for example comes with 1.8.3, which doesn't support this.

PR #683, opened 2019-05-06
glib-2.0: gdbus: Avoid printing null strings

MastaG:
Backport fix so we can build again on Fedora 30
See: https://gitlab.gnome.org/GNOME/glib/commit/566e1d61a500267c7849ad0b2552feec9c9a29a6


WanWizard:
Completely missed the release announcement.

MastaG:
Well it's a very nice release with gcc 9!
Phoronix always writes nice articles about these things:
https://www.phoronix.com/scan.php?page=news_item&px=Fedora-30-Released

PR #684, opened 2019-05-07
mdev-mount - use 'lazy' switch for umount

ims21:


littlesat:
When you are connected from PC to this pluged usb stick in box (f.eg. /media/SATS) and stick is removed, then system on e2 cannot umount it due "busy" => directory SATS is not removed => next plug-in this stick to box is not as /media/SATS but as /media/sdxx  (xx = c1, d1 ...)

With -l it is umounted well and empty directory is removed => next plug is ok.

PR #685, opened 2019-05-09
dos2unix files

persianpros:


atvcaptain:
patch files dont change it its a bad idea

WanWizard:
I think that happened in a previous one too?

persianpros:
No, I think rebuilding of packages because of patch changes is what he means.

WanWizard:
I can live with that. 

As long as the patches apply correctly, I assume CRLF line ends were there because they were in the source, and therefore ended up in the diff. Which means it would have been correct for the CRLF to be there.

PR #686, opened 2019-05-13
rtl8812au: Fix build with kernel 5.1

persianpros:


WanWizard:
Doesn't fix the build:
> ERROR: rtl8812au-5.1.5-r0 do_patch: Command Error: 'quilt --quiltrc /openpli/develop/build/tmp/work/vuultimo4k-oe-linux-gnueabi/rtl8812au/5.1.5-r0/recipe-sysroot-native/etc/quiltrc push' exited with 0  Output:
Applying patch 0001-add-linux-kernel-5.1-support.patch
patching file os_dep/osdep_service.c
Hunk #1 FAILED at 1974 (different line endings).
Hunk #2 FAILED at 2004 (different line endings).
Hunk #3 FAILED at 2039 (different line endings).

The source file has CRLF line ends...

persianpros:
Maybe my windows 10/notepad++ did something to the file as I don't always use ubuntu, this is the original file: https://github.com/oe-alliance/oe-alliance-core/blob/4.2/meta-oe/recipes-connectivity/realtek/rt8812au/0001-add-linux-kernel-5.1-support.patch

WanWizard:
That is indeed it. The other ones are needed as well, btw. I'll add them.

PR #687, opened 2019-05-15
DLNAServer & ReadyMedia

MichaelSweden:
Main reason for this change was that I missed a setting in the minidlna (ReadyMedia) and while running with the DLNAServer plugin it wasn't possible to manually change this. I have added this new setting (root container) to the plugin. This will give less folders for user at the TV. I also wanted to use one media folder, it was always three via the plugin. During testing I found some bugs and fix them. Because of bigger change of the plugin code, I realize that continue with patching this old code isn't a good idea. So I made a copy and now the plugin source code is in our repo. The user functionality Reset is changed. Now it will reset settings back to hardcoded decent values. If a harddisc is accessible it will use the Movie folder, else Media. I also made it possible to edit server name.

The ReadyMedia took the source from an old not updated repo. The origin repo is at sourceforge, but I take from a mirror at github. I reduce the huge sql cache and change some default config.

I have removed the minidlna module, since it is now called readymedia (and this is used by the plugin). Note however that a minidlna is available in the meta-openembedded.

To setup a DLNA Server:
The user should install the DLNAServer plugin. Readymedia (minidlna daemon) will be installed automatically. Then the user should goto the plugin. In the dialog is recommended (first time) to press Reset, Save, Start. Of course if wanted settings can be changed before Save/Start.


WanWizard:
I'll leave this open for the moment if you don't mind. We're in the middle of a release candidate, I don't want to commit stuff to develop untl the rc branch is stable, to avoid merge issues.

MichaelSweden:
No problem, no hurry. Please take it after 7.1 release.

persianpros:
@MichaelSweden Please send a PR for your "01-less-sql-page-cache.patch" and I will merge it ASAP, also now we have "minidlna.conf" and "readymedia.sh" for enigma2 so update your bitbake recipe please.

meta-openpli/recipes-multimedia/readymedia/readymedia.bb
```diff
@@ -8,12 +8,12 @@ LIC_FILES_CHKSUM = "file://LICENCE.miniupnpd;md5=b0dabf9d8e0f871554e309d62ead8d2
 
 inherit gitpkgv
 
-PV = "1.1.0+git${SRCPV}"
-PKGV = "1.1.0+git${GITPKGV}"
-PR = "r0"
+PV = "1.1.6+git${SRCPV}"
```
persianpros: It's "1.2.2" now ;)

MichaelSweden: The number I wrote in PR isn't the version of minidlna. I didn't know what to write, but found some numbers in git and took it. I only wanted to update it. It takes latest commit, so the recipe doesn't now the version. I am new to bitbake and this build environment. I have now figure out that it is enough to change PR or PV. So in this case since I update PR, no requirement to update PV. However nothing wrong to update both.

Can I keep it as it is or what should I write (I don't think it should be related to version of minidlna, but maybe it can be, but it is confusing)?

MichaelSweden:
What is the intention with the OpenVisionE2/minidlna repo? Will it always try to mirror the originate repo, or diverge it's own way? Will it be focus on STB and not PC servers? Maybe it would be a good idea to write information about this on github page.

I have now tested and take a closer look at this (the change of sql cache). I actually cannot see any change in memory usage (also tested in a Ubuntu machine). I have talked to a guy that have experiance with sqlite, however he never have changed any cache and didn't know about this. I will later on take a deeper look and try to figure out what is going on. If this is good to have I can request a change to the minidlna, as you suggested.

I have now seen your changes to minidlna, I can adjust.

persianpros:
OpenVisionE2/minidlna has updates that original repo doesn't have and the original repo won't get updates after 2018.

Our focus is on STB but you could create a pc branch for it if you want.

meta-openpli/recipes-multimedia/readymedia/readymedia.bb
persianpros: There's no need for PR, just remove it and keep PV with our version which is now 1.2.2 as AUTOREV will get latest updates.

This recipe needs some changes, MAINTAINER is us not Open PLi, there's no need for external "readymedia.sh" and "minidlna.conf" as we provide them.

MichaelSweden:
Thanks for information. I wonder if you have thought about the name. Should you call it MiniDLNA or ReadyMedia? I cannot all history and I haven't found a good reason for the ReadyMedia name. What does it actually stand for, I mean why a new name? In OpenPLi this name (I think) is mostly used because it is used by DLNAServer plugin. Why not call it minidlna, since still files are called that. It's confusing with two names. So why not "MiniDLNA for STB". I only come with suggestions and also I would like to understand why it is as it is. Maybe you have thought and discuss this before. So little about ideas that have pop up in my mind... Skip the ReadyMedia name. Update the README file (maybe a nice README.md). Write something about it is mainly for Enigma2 based STB. Keep compatibility with PC machines. Rename readymedia.sh to minidlna.sh, write comment in the file that it is for STB use (for others a minidlna file is created from minidlna.init.d.script.tmpl). Maybe can write how install is done on STB (since the normal install isn't used), this might belong to the readme file. Update default stop to 0 1 6 (however this isn't used from my recipe). Since it is no default stop, maybe that should be commented. - Just a few kind thoughts from me.

I'm happy to see progress in this software. I have idea about a new functionality and now your repo seems to be the ground for it. I think it should keep existing functionality/install and so on for PC machines (e.q. Ubuntu). This for easy development and also normal use. I have also found some issues which I might take a closer look at in the future.

I have little time right now, but will of course update my recipe.

persianpros:
I vote for "minidlna" name nothing more :+1: 

I'm happy too that you're working on this feature, we have a nice readme for our image and you can use it as your sample: https://github.com/OpenVisionE2/openvision-oe/blob/develop/README.md 

I just sent you invitation link with write access to our repo.

persianpros:
@WanWizard would you please squash merge this into develop so we could use it?

meta-openpli/recipes-multimedia/readymedia/readymedia.bb
```diff
@@ -1,23 +1,18 @@
-DESCRIPTION = "ReadyMedia DLNA server (minidlna)"
+DESCRIPTION = "MiniDLNA a DLNA/UPnP-AV server"
 SUMMARY = "Please use the DLNAServer plugin, enigma2-plugin-extensions-dlnaserver."
-HOMEPAGE = "http://sourceforge.net/projects/minidlna"
-SECTION = "multimedia"
-MAINTAINER = "OpenPLi"
+HOMEPAGE = "http://github.com/OpenVisionE2/minidlna"
+SECTION = "network"
+MAINTAINER = "OpenVisionE2"
```
persianpros: This should be like https://github.com/OpenPLi/openpli-oe-core/blob/develop/meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-backupsuite.bb#L3

I mean "Open Vision Developers" ;)

littlesat:
This will be evaluated as soon rc becomes release... so please hold-on

PR #727, opened 2019-08-23
[bootlogo.sh] Add option for skin bootlogo

IanSav:
This change adds the option for skins to include a "bootlogo.mvi" image at the top level of the skin that will be used when that skin is selected.

The code first checks to see if a skin is defined in the "settings" file and uses it if found. If no skin is identified then the "skin.pyo" file is searched to see if a built-in skin has been defined. If found then that will be used. If no skin associated bootlogo is identified then the code will progress to the original search in "/etc/enigma2/" and finally use the image in "/usr/share/".


IanSav:
This proposed change is not acceptable to all images.  I will revert to the previous mode of operation.

PR #728, opened 2019-08-26
Don't hardcode path

persianpros:


WanWizard:
Quite a few of these recipes are backported from newer OE versions, so you might have to do this upstream too, otherwise it will be undone the next time something is backported.

PR #737, opened 2019-11-23
Thud qt

zgemma-star:
qtwebkit and qtwebengine for upgrading now and in the future it is easier to use the qt.io git repository.
the qt.io is better up-to-date which is prefer for DRM content websites which some carefully checks if the platform has the latest security upgrades. 
when the new branch is accept then it can been update to 5.12.6 which users in qt see at the moment one of best stability version at the moment.


betacentauri:
Have you tested it? Does everything work?
I need some days to be able to test it as build server currently build all images from scratch.

zgemma-star:
Yes, build and test on Hisilicon and Broadcom chipsets.

PR #766, opened 2020-02-21
Import upstream Wiregard 0.0.20191219 from meta-openembedded

MastaG:
Just import them but not ship it yet in the feeds since I still need to go through all the defconfigs.


athoik:
Newer recipes commited here https://github.com/OpenPLi/openpli-oe-core/commit/0e8753e74445aa1015f4f4bd6b9053712b5da05c

PR #773, opened 2020-03-19
Add Glamour Aura Color skins to the feeds

MCelliotG:
- New Glamour Aura Color skins, directly from my Github (master branches)


MCelliotG:
The skins are available on my Github
https://github.com/MCelliotG/GlamourAuraBerry-skin
https://github.com/MCelliotG/GlamourAuraLime-skin
https://github.com/MCelliotG/GlamourAuraMango-skin
https://github.com/MCelliotG/GlamourAuraRose-skin

WanWizard:
For those to be build, you also need build recipes. I'll have a look at those, and keep this PR open as a reminder.

MCelliotG:
Thank you! Is there any tutorial on how to build recipes btw?

WanWizard:
This is an example of your first one: https://github.com/OpenPLi/openpli-oe-core/blob/develop/meta-openpli/recipes-openpli/enigma2-skins/enigma2-plugin-skins-glamouraurafhd.bb

MCelliotG:
And how are the control files determined with this code? Having the correct control files in the installer is crucial for the Color skins. Are they retrieved automatically from the CONTROL folder I provide?
So a code like this is enough for instance?
````
DESCRIPTION = "Glamour Aura Berry skin for new generation STBs with OpenPLI based images"
MAINTAINER = "MCelliotG"
LICENSE = "GPLv3"
LIC_FILES_CHKSUM = "file://LICENSE;md5=84dcc94da3adb52b53ae4fa38fe49e5d"

inherit gitpkgv allarch

PV = "6.0.5+git${SRCPV}"
PKGV = "6.0.5+git${GITPKGV}"

SRC_URI = "git://github.com/MCelliotG/GlamourAuraBerry-skin.git"

FILES_${PN} = "/usr"

S = "${WORKDIR}/git"

do_compile() {
}

do_install() {
	install -d ${D}/usr
	cp -r --preserve=mode,links ${S}/usr/* ${D}/usr/
	chmod -R a+rX ${D}/usr
}
````

WanWizard:
No,control files are generated as part of the build process. We don't do anything hardcoded. If the recipe is correct, so will the control file be.

Assuming the new skin has the same layout and the other one, the recipe should be fine. If 6.0.5. is correct as your version number.

Once the recipe is finished and tested, the final step is to add it here: https://github.com/OpenPLi/openpli-oe-core/blob/develop/meta-openpli/recipes-openpli/images/openpli-enigma2-feed.bb#L188, so it will be added to the feed.

MCelliotG:
The problem is that all common files of the color skins are symlinked to the main GlamourAuraFHD skin, and the control files check if the main GAFHD skin is installed and if not, they install it as a dependency and symlink silently the common files to the main skin. This way all skins work seamlessly and the file size of the color skin is reduced significantly. Now, the control file of the main skin was also modified to check if additional Color skins are installed and re-symlinks the common files. This way if I add an icon or new common file to the main skin, the color skins are updated as well with every update of the main skin! 
I think there must be a way for a recipe to feed from a control file, otherwise how other packages with dependencies are built?
Sorry for the many questions, I still learn about these things!

WanWizard:
What you say is that when a user installs this skin, the system has to make sure the main skin is installed first?

You do that by defining a dependency in the recipe:
````
RDEPENDS_${PN} += "enigma2-plugin-skins-glamouraurafhd"
````

I wonder if symlinking is a good idea. It is definitely not a good idea to add code to one package that updates another package, both are things that opkg definitely doesn't like, it messes with its version management.

@IanSav, you know a lot more of the skin system than I do, what is the best way to have multiple skins share resources? If at all possible?

MCelliotG:
Yes, I created the skins this way to reduce file sizes. In the standalone IPKs I make I include the correct control files and these make sure to check if the main skin is installed. I can understand that control files are auto generated but there must be a recipe to make sure the control files are updated too from the github source.
Here's an example from one of the color skins
**postinst**
`#!/bin/sh
if [ -f /tmp/GlamourAuraBerry/skin.xml ]; then
	rm -rf /usr/share/enigma2/GlamourAuraBerry/
	mv /tmp/GlamourAuraBerry /usr/share/enigma2/
fi
echo "                                                                                 "
echo "Glamour Aura Berry main files installed, now linking common files to main skin..." 
cp -Rs /usr/share/enigma2/GlamourAuraFHD/* /usr/share/enigma2/GlamourAuraBerry 2>/dev/null
echo "                                                                              "
echo "Glamour Aura Berry skin is now successfully installed on your receiver        "
echo "                         Enjoy!!!                                             "
`
this check links the common files, it works like a charm with no problems.

And here's the check from the main skin
**postinst**
`#!/bin/sh
if [ -f /tmp/GlamourAuraFHD/skin.xml ]; then
  rm -rf /usr/share/enigma2/GlamourAuraFHD/
  mv /tmp/GlamourAuraFHD /usr/share/enigma2/
fi
echo "                                                                     "
echo "Now, checking if additional Aura Color skins are already installed   " 
if [ -f /usr/share/enigma2/GlamourAuraBerry/skin.xml ] || [ -f /usr/share/enigma2/GlamourAuraLime/skin.xml ] || [ -f /usr/share/enigma2/GlamourAuraMango/skin.xml ] || [ -f /usr/share/enigma2/GlamourAuraRose/skin.xml ]
then
  echo "Glamour Aura Color skins, found, now updating links to main skin..." 
  echo "                                                                   "
else
  echo "Glamour Aura Color skins not found, no file linking necessary      " 
  echo "                                                                   " 
fi
if [ -f /usr/share/enigma2/GlamourAuraBerry/skin.xml ]; then
  cp -Rs /usr/share/enigma2/GlamourAuraFHD/* /usr/share/enigma2/GlamourAuraBerry 2>/dev/null
fi
if [ -f /usr/share/enigma2/GlamourAuraLime/skin.xml ]; then
  cp -Rs /usr/share/enigma2/GlamourAuraFHD/* /usr/share/enigma2/GlamourAuraLime 2>/dev/null
fi
if [ -f /usr/share/enigma2/GlamourAuraMango/skin.xml ]; then
  cp -Rs /usr/share/enigma2/GlamourAuraFHD/* /usr/share/enigma2/GlamourAuraMango 2>/dev/null
fi
if [ -f /usr/share/enigma2/GlamourAuraRose/skin.xml ]; then
  cp -Rs /usr/share/enigma2/GlamourAuraFHD/* /usr/share/enigma2/GlamourAuraRose 2>/dev/null
fi
echo "                                                                    "
echo "Glamour Aura FHD skin is now successfully installed on your receiver"
echo "   `

Same way if any of the color skins are uninstalled, the rest remain completely unaffected.

MCelliotG:
Also on OpenATV when they also made the recipe for the version of my skin for their images (other github repositories since I maintain both images) they used my control files as they were, so I believe there must be a way in the recipe to indicate where to use the control files.

WanWizard:
ATV has a habit of hacking their way through, that is entirely up to them. They also don't do full bitbake build runs like we do, and if they use your control file, they don't even use a recipe to build the ipk file, they build it by hand ( = using code instead of bitbake), which is even worse.

We like to keep it clean, if one of the hacks fall, our entire buildrun fails, and we can't afford that.

Those pre- and postinst sections are completely unacceptable for us.

MCelliotG:
Hmmm, I understand. I'm just curious, how can skins use shared resources? Otherwise I will have to make changes on my github, but I think it's doable. I think it is worth a check on how to!

MCelliotG:
I found something here, don't know if that applies to this case
https://dominion.thruhere.net/koen/OE/OE-usermanual.html

WanWizard:
The perferred OpenEmbedded way is to create a "common" package, that contains all files that are shared between the different skins, and add a dependency from all skin packages to this common package. opkg will then make sure the common package is installed when you install the first skin, others will use it when it's there, and opkg will remove it only after the last skin that depends on it is removed.

That would require reorganizing of files, but as said I don't know if the current skin system has an easy solution for this kind of setups (if not it should imho), hence my question to @IanSav

MCelliotG:
I tried to work on github with the symlinks as files, but that solution failed. So, I updated my Github with the regular files that were initially symlinked. Now, the skins should build normally, as long as there is an rdepends command in the recipe pointing to the main skin.
In the near future, I will try to change the paths in xml pointing to any common file possible. Unfortunately that is impossible for files that are hardcoded (like for instance the movie list icons and others), but I don't think these take up that much space.
So, please test if the new skins build OK (they should) and let me know!
Thanks in advance!

WanWizard:
That is indeed a workaround, but means file duplication, and with that, more maintenance. So ideally this should be avoided.

MCelliotG:
Yes, but pointing new paths in 5x3=15 xml files for OpenPLI (and other 15 for openATV) will take some time! I will do it of course, since there seems not to be any other way, if building a custom postinst script in the feeds is no option. Unless you suggest I wait if this solution is possible.
Having a common folder will not work in any way, what will work if the cp - Rs command cannot run after the color skin installation is to point any file that can be shared that is not required in a hardcoded location. This can happen gradually, meaning that with each skin update, I will remove any file that is duplicate, and the final skin sizes will shrink.

MCelliotG:
BTW, I just found this
https://stackoverflow.com/questions/28397650/bitbake-recipe-to-have-pre-and-post-install-action
https://www.yoctoproject.org/docs/1.7.1/mega-manual/mega-manual.html

And especially this
`5.3.16. Post-Installation Scripts¶

Post-installation scripts run immediately after installing a package on the target or during image creation when a package is included in an image. To add a post-installation script to a package, add a pkg_postinst_PACKAGENAME() function to the recipe file (.bb) and replace PACKAGENAME with the name of the package you want to attach to the postinst script. To apply the post-installation script to the main package for the recipe, which is usually what is required, specify ${PN} in place of PACKAGENAME.

A post-installation function has the following structure:

     pkg_postinst_PACKAGENAME() {
     #!/bin/sh -e
     # Commands to carry out
     }
                

The script defined in the post-installation function is called when the root filesystem is created. If the script succeeds, the package is marked as installed. If the script fails, the package is marked as unpacked and the script is executed when the image boots again.

Sometimes it is necessary for the execution of a post-installation script to be delayed until the first boot. For example, the script might need to be executed on the device itself. To delay script execution until boot time, use the following structure in the post-installation script:

     pkg_postinst_PACKAGENAME() {
     #!/bin/sh -e
     if [ x"$D" = "x" ]; then
          # Actions to carry out on the device go here
     else
          exit 1
     fi
     }
                `

WanWizard:
There is no issue with scripts as such, many packages use them. 

The problem is with dependencies, scripts requiring other packages to be installed, to be up to date, and to be installed in the correct sequence. This should not be dealt with in scripts, this is why we have a package manager.

The main question here is "how can I share a resource amongst multiple skins", which I can't answer as it is not my area of expertise. @IanSav is, so I'm awaiting his response.

MCelliotG:
But I don't have any other dependencies other than the main skin itself, which is already in the feeds. Even the local ipk tests require that the main skin is installed first in that sequence.

WanWizard:
And what happens if I decide to deinstall the main skin because I don't like it, but still want to use one of your other skins?

IanSav:
> @IanSav, you know a lot more of the skin system than I do, what is the best way to have multiple skins share resources? If at all possible?

Sorry for the late reply.  I have been busy on an OpenATV issue.

It is very bad for one package to try to manipulate files in another package.

One way to have a complex skin with different colour options is have the entire bulk of the skin as an included file or files.  You then create a series of different skins that have all the colour specifications that then use the common skin XML file / files for all the screen layout definitions.  You then bundle the whole lot into a single package.  You install one skin and you get all the colour options.  Because the multiple skins are using mostly common XML the size should not be a problem.

The other way to have multiple colour options in a skin is to have a single skin with a plugin that manager that skin.  An example of that approach is my OverlayHD skin.  There is only one set of skin code XML files and an associated plugin that changes the colours and some other configurable options within the skin.  It could be said that OverlayHD is a skin with almost infinite colour combination options.

MCelliotG:
You guys made me even more confused, I don't even know what to say anymore...
I may remind you that while I considered all these options, I am NOT a coder, so if I knew the way to create a plugin, I would have done that already. Even so, all skins I know using plugins they simply change the color hex code of their xmls, they don't colorize graphics. I may be mistaken, of course.
Now, I don't understand how a simple postinst command will manipulate files, it leaves the main skin completely untouched. Since we don't want duplicate files, we need dependencies, I don't see any other way. The color skins come as an extra to the main skin, that was my initial idea. The other one will be to create full independent standalone skins, but that means more files, more size, and even separate python system, unless we make the pythons a dependency.

So let's revise my options (the ones I can actually do)
1. make a recipe with a postinst script that create symlinks to the extra skins (I already provide that) and leave the rest of the structure on the extra skins minimal. The main skin is still required.
2. Make completely independent skins. This will make large files, so if someone wants all 5 skins in their receiver, they will take up a lot of space. But it will free users from requiring the main skin as a dependency.
3. Create a package with all common files as a dependency (python files and common graphics). This will make users choose between any of my skins.
4. Modify the xmls of the color skins, in order to point to the common files of the main skin. Also needs the main skin as a dependency.
5. Completely abandon the project I worked so hard on for many months. :)

So, please tell me what to do. In my logic, option 1 is what is feasible right now, as it requires no more extra work from me. All other options will delay the launch significantly!

IanSav:
I didn't intend to confuse you.  I haven't looked closely at your skin.  I am simply responding to WanWizard's  request for my comment.

You are quite correct that most skins, including OverlayHD, play with colours and not images.  OverlayHD's plugin could easily be expanded to include swapping image libraries.

I am quite busy at the moment but I would suggest you complete your efforts.  Just keep in mind that no single package should interfere with the contents of any other package.  In this case all your alternate colour option packages should have dependencies on the core skin package.  I would also like to ask that you try to avoid scattering the skin's python components into the Converter and Renderer directories.  I am also concerned about the use of a "skin_default" subdirectory within the skin, this is likely to create issues.

When your work is complete feel free to contact me and I will try to find time to code review your work.  I must ask and insist that if the finished product changes *any* standard files within the image that you tell me as that will stop me from getting involved as I will not compromise my development environment.

MCelliotG:
Thanks for the input.
I made up my mind and change the paths of the color skins to point out to the main skin. If you take time to understand the concept, color skins are extensions to the main skin, so they won't exist separately. This is what I'm going for from the beginning.
The rest of the critique, although general and I get it, sounds like my skin would cause issues, when it clearly does not. The skin_default folder is there to ensure every hardcoded png that cannot be defined in paths in the xml will be skinned as well, I don't want to use any default icons from the main system, it messes up my work's aesthetic completely. These are issues well put behind and tested not to mess anything in the enigma2 system. That surely includes the skin's python system. Absolutely **nothing** is modified in Enigma2's system, and that's **my main principal** also. Everything in the skin runs independently from installation to removal, and you can make sure of that.
Anyway, guys thank you so much for the input and your valuable suggestions and above all your time, I really appreciate it.
Now, l have a lot of work to do....

littlesat:
when a skin shares parts then make a recipe for the shared part and add depends on the skin for that specific shared part.....? I understand it could be annoying doing It right... I agree even with our standard skin this should be improved

PR #782, opened 2020-04-16
Thud kodi18

HD-Digital:
Test with HD51 and HD61
 
todo: Kodi18 need ffmpeg => 4 (now 3.4.2 is preferred)
todo: Kodi addons test
todo: BSP layer update


littlesat:
Do we
Need to wait for the bsp layer before merging?

HD-Digital:
Only for broadcom it needed An BSP update for the v3d drivers 

Ffmpeg it need version 4 but I didn't want to force to remove 3.4.2 on some way this is preferred.

HD-Digital:
goes further here: https://github.com/OpenPLi/openpli-oe-core/pull/787

PR #790, opened 2020-04-27
ffmpeg 4.% add codec support (probably need some fine

HD-Digital:
tune)
ffmpeg 3.4.2 change to ffmpeg-old (for if you prefer 3.4.2 keep for now available)


athoik:
I prefer to move on newer ffmpeg. Thanks!

PR #795, opened 2020-05-01
python package update

HD-Digital:
cause issues like 
 ( :0: UserWarning: You do not have… a working installation of the service_identity module: 'cannot import name verify_ip_address'.  Please install it from <https://pypi.python.org/pypi/service_identity> and make sure all of its dependencies are satisfied.  Without the service_identity module, Twisted can perform only rudimentary TLS client hostname verification.  Many valid certificate/hostname mappings may be rejected. )


athoik:
Hi,

isodate 0.7 is not released yet!

```
ERROR: python-isodate-0.7.0-r0 do_fetch: Fetcher failure for URL: 'https://files.pythonhosted.org/packages/source/i/isodate/isodate-0.7.0.tar.gz'. Unable to fetch URL from any source.
```

It doesn't contain also any changes since 0.6

https://github.com/gweis/isodate/commits/master

We have to revert isodate back to 0.6.

PR #807, opened 2020-05-26
add dvb usb xbox one tuner

zgemma-star:
Test and xbox one tuner was found
https://github.com/zgemma-star/meta-zgemma/commit/85a9facb4780e9de8730b4cfb496be013cb6f88d


WanWizard:
Not sure this is a good idea.

This will create a package that is useless on kernel versions that don't deliver those drivers, due to the RRECOMMENDS, which is why we haven't added it yet.

/cc @eriksl

WanWizard:
Brillant, thanks a million!

zgemma-star:
On some reason a small part was already pushed. It wasn't yet complete :) 
https://github.com/OpenPLi/openpli-oe-core/pull/808

for add backport devices in the bsp layer: https://github.com/zgemma-star/meta-zgemma/commit/35dca9643f98d45e70ef9b4b04a7e8069a3afe39#diff-5e99acf204dec63948824b9e7f15f68aR52

PR #828, opened 2020-06-20
rtl8821cu: wifi drivers for rtl8811cu and rtl8821cu

wte-chromium:


wte-chromium:
Drivers will not build on dm8000 and xtrend with kernel 3.8.x, so from kernel 4.0 it seems to build fine.
Edision kernel 5.5.16 didn't support ".policy" so for now increase kernel version for it to 5.6 or higher

PR #829, opened 2020-06-21
wifi rtl88x2bu: use original realtek package

wte-chromium:
(better bluetooth support and better wifi speed)


wte-chromium:
We need to set some models of xtrend to the older rtl88x2bu packages as the original/newer packages don't support older kernel support anymore.
https://github.com/neo-pli-bsps/meta-xtrend/pull/4

PR #841, opened 2020-07-11
Don't use AUTOREV for BackupSuite

persianpros:
https://github.com/OpenVisionE2/BackupSuite/commit/276a02172ff872cf1069d55ee0248ccd94cc2274 isn't compatible with PLi as we're using full machine names like "vuduo" instead of "duo" in the new "lookuptable.txt"

https://github.com/OpenVisionE2/BackupSuite/blob/master/plugin/backupsuite.sh#L214 returns "duo" for you so the new "lookuptable.txt" won't understand it.

Is there a way to detect machine names like getBoxType but in shell not python? (For PLi I mean as we're fine)

### DIFF
--- a/meta-openpli/conf/distro/reporefs.conf
+++ b/meta-openpli/conf/distro/reporefs.conf
@@ -189,7 +189,7 @@ SRCREV_pn-minisatip = "${AUTOREV}"
 # plugins
 SRCREV_pn-enigma2-pliplugins ??= "${AUTOREV}"
 SRCREV_pn-enigma2-plugin-extensions-automatic-fullbackup ??= "${AUTOREV}"
-SRCREV_pn-enigma2-plugin-extensions-backupsuite ??= "${AUTOREV}"
+SRCREV_pn-enigma2-plugin-extensions-backupsuite ??= "c7f27f896824ce6fc40fb2fcd852a297373cb330"
 SRCREV_pn-enigma2-plugin-extensions-blurayplayer ??= "${AUTOREV}"
 SRCREV_pn-enigma2-plugin-extensions-dlnabrowser ??= "${AUTOREV}"
 SRCREV_pn-enigma2-plugin-extensions-dlnaserver ??= "${AUTOREV}"


WanWizard:
I don't think it is entirely correct that a generic plugin who's maintenance you have inherited is now changed to work on your own image only. And leave it to other images to fix the problem you have created.

persianpros:
If you check the last commits I tried to keep it compatible with all images not just Open Vision and even support all brands/models which were not there.

The latest commit solves box detection problems on Open Vision but makes a new problem for PLi that's why I pointed to a specific commit hash for PLi so I find a way to fix it for all, I'm asking for a way to detect for all not making this plugin for OV only.

We store machie name in "/etc/openvision/model" so we could use it in shell and other places which we can't use getBoxType, do you keep machine names somewhere?

persianpros:
I can provide a kernel module to create

/proc/stb/info/enigma2model

for example to store the MACHINE name from OE, if you're agree with this we could use it instead of

/proc/stb/info/hwmodel
/proc/stb/info/gbmodel
/proc/stb/info/boxtype
/proc/stb/info/vumodel
/proc/stb/info/azmodel

to check one proc file globally for all in all places, even reduce checks in enigma2 itself.

WanWizard:
I suggest to have this discussion in the forum. I am not a developer, this isn't up to me.

Personally I am still a proponent of a generic and universal "boxid" module that can be used by all images, so we can avoid problems like this.

persianpros:
Done: https://forums.openpli.org/topic/77494-one-proc-file-for-detecting-the-machine-in-all-enigma2-images/

Yes exactly a global module for this will fix so many problems and reduce the checks everywhere.

WanWizard:
Thanks. I've left my 2ct's there. ;-)

persianpros:
https://github.com/OpenPLi/openpli-oe-core/pull/842 will do it for PLi if you're up for it.

persianpros:
I'm not sure if I need to change something after using udev instead of mdev ...

persianpros:
Ok I created a "develop" branch for Open Vision as we have our own branding module (https://github.com/OpenVisionE2/branding-module) and we could check so many other things (even more than OE-A).

So master branch is safe for you and we have our own branch.

Two of PLi members have access to BackupSuite if you or any other PLi dev want access let me know :+1:

PR #842, opened 2020-07-16
Introduce enigma2-module to provide MACHINE name in /proc/stb/info/enigma2model

persianpros:
https://forums.openpli.org/topic/77494-one-proc-file-for-detecting-the-machine-in-all-enigma2-images/

### DIFF
--- a/meta-openpli/recipes-openpli/enigma2-module/enigma2-module.bb
+++ b/meta-openpli/recipes-openpli/enigma2-module/enigma2-module.bb
@@ -1,0 +1,33 @@
+SECTION = "kernel/modules"
+LICENSE = "CLOSED"
+PRIORITY = "required"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+SRC_URI = "file://enigma2module.c file://Makefile"
+
+S = "${WORKDIR}"
+
+inherit module machine_kernel_pr
+
+EXTRA_OEMAKE = "KSRC=${STAGING_KERNEL_BUILDDIR}"
+
+do_compile() {
+	unset CFLAGS CPPFLAGS CXXFLAGS LDFLAGS
+	oe_runmake -C "${STAGING_KERNEL_BUILDDIR}" M="${S}" modules
+}
+
+do_configure[nostamp] = "1"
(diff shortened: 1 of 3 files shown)


zgemma-star:
/proc/stb is driver part. We have already boxtype available as setting, this will in next driver been improved.

Please don't make a mess with this kind of code or just like osrelease and so on place it in /etc

zgemma-star:
The driver name is in openpli machine name, so what do you gain with it?

WanWizard:
There is a discussion about this on the forum. 

It's basically a workaround for missing boxbranding in OpenPLi, which should provide a unified interface layer to query the hardware, without having to deal with all the complexities and differences in runtime code.

persianpros:
@zgemma-star 

There isn't any standard yet and a path isn't anything special.

68foxboris:
root@et1x000:~# cat /proc/stb/info/enigma2model
et1x000

root@et1x000:~# cat /proc/stb/info/model
dm800se

zgemma-star:
First of all why use /proc/stb as this is generate by drivers. It's better to use a separate map like /proc/boxbranding
Second I see no advantage in this because machine/driver name is already provide in /proc/stb/info/model
I would advice to inform manufactures and fix it who use model name dm800se. It's easier in the end to make a small work-a-round for this "dm800se" boxes then make work-a-rounds in all images.

In the next driver release /proc/stb/info/boxtype will inform the type of box like H9 Twin, H9s and so on. I think only advance for this is that you can show other remote and frontpanel design under help button or owif.

zgemma-star:
I am not against to add on some way identificiation but it must been clear if it's part of driver, images, enigma2, openembedded and so on.
Why not just use os-release and add machine info to it for the few boxes which send wrong info back...
https://git.yoctoproject.org/cgit.cgi/poky/plain/meta/recipes-core/os-release/os-release.bb

68foxboris:
Unfortunately for ET 110004K nothing can be done. There is no support for a very long time.

persianpros:
@68foxboris 

Most of times /proc/stb/info/model is fake.

PR #856, opened 2020-08-29
remove old patches

kueken:
included new branch dvbmediasiink


athoik (commented):
Those patches required for dreambox.

If you are dropping support for dm8k it's ok to remove them.

kueken:
OK,
I will close the pull request and test the build via pli-mediasink, but that takes time.
You can certainly change it easily.

PR #857, opened 2020-08-29
fix warning

kueken:


kueken:
back to pli-dvbmediasink

As requested.
@athoik 
Perhaps the comment is allowed ...
However, I cannot understand why the patches are not integrated into your 'own' pli-dvbmediasink '?
I hope that's okay.
greetings

WanWizard:
if you refer to the dmm patch, we don't do "if boxname then" in our code, if we can avoid it.

kueken:
OK,
I've built a couple of images with no errors.
Hisi / mipsel and everything seems fine.
Hopefully the rest of the build works for you without any errors, too.
greetings

WanWizard:
I've started a full build run.

athoik:
The recipe also include new changes like the following.

```
# added to have al m4 macro's into build when using bitbake with -b option.
# Then proceeding to full image build or at least package build with recipes parsing is not needed.
do_configure_prepend() {
    ln -sf ${STAGING_DATADIR_NATIVE}/aclocal/*.m4 ${S}/m4/
}

do_configure_prepend() {
    sed -i 's/AC_INIT.*$/AC_INIT(gst-plugin-dvbmediasink, 1.0.0, @pli4)/' ${S}/configure.ac
    sed -i 's/AM_INIT_AUTOMAKE.*$/AM_INIT_AUTOMAKE([foreign])/' ${S}/configure.ac
    touch ${S}/NEWS
    touch ${S}/README
    touch ${S}/AUTHORS
    touch ${S}/ChangeLog
}
```

Also the following:

```
pkg_preinst_${PN}_prepend () {
	if [ -d "/.cache/gstreamer-1.0" ]
	then
		rm -rf "/.cache/gstreamer-1.0"
	fi
	if [ -d "/home/root/.cache/gstreamer-1.0" ]
	then
		rm -rf "/home/root/.cache/gstreamer-1.0"
	fi
}
```

In case we would like to keep those in the bitbake, it would be great to explain with a comment or on the commit why are adding those special instructions on the bitbake. Only do_configure_prepend contains a useful comment. 

As OpenPLi team, long time ago we aborted CVR dvbmediasink, because it was introducing lot's of IF BOX code.

If nowadays OpenPLi likes to switch to CVR dvbmediasink, personally I don't mind, but I would like to know what are the pros/cons switching to CVR repository.

WanWizard:
I didn't notice the switch initially, and I don't have the knowledge to have an opinion.

kueken:
@athoik 
I honestly don't understand your objections.
But you still have direct access to the OE build.
I just wanted to update media framework.
The construction is flawless and it works too, but the structure doesn't seem to please.
You are welcome to adapt and / or optimize it.
I probably have less idea than you do.
Please adjust it according to your wishes.
Just wanted to contribute a little.
Best wishes

athoik:
There is different structure in OE-Alliance and in OpenPLi.

Also decisions taken long time ago not to use some stuff, like CVR fork of dvbmediasink.

So it should be clear what changes we are applying to recipes, other than changing version revisions.

kueken:
I have no idea about the internal disputes between Pli and OE-Alliance.
I don't care either.
If there are any problems just tweak things.
greetings

PR #873, opened 2020-11-05
Update 3rd-party-feed-configs.bb

persianpros:
Am I the only one seeing this?


WanWizard:
Appearantly because there is no error in the end result?
````
root@vuduo4k:~# cat /etc/opkg/3rd-party-vuduo4k-feed.conf 
src/gz openpli-3rd-party-vuduo4k http://downloads.openpli.org/feeds/openpli-7-release/3rd-party-vuduo4k
````

PR #875, opened 2020-11-07
update youtube_dl again

kueken:
github is also possible again
add ca-certificate and codec rdepends (not sure)


WanWizard:
ca-certificates is installed by default, not sure why python-codecs would be needed?

Also, why a new URL, what is wrong with the current one?

kueken:
depends I'm not really sure ...
Url with previous address was no longer updated.
The new Url is also the current link from git.
Is also the latest version with a new readme
https://github.com/l1ving/youtube-dl

kueken:
update and remove new depends

PR #918, opened 2021-02-19
fakelocale: add Macedonian

blzr:


blzr:
New PR with more unambiguous description will be submitted soon...

PR #920, opened 2021-02-23
update vuplus fix scr

kueken:


WanWizard:
We have an automated script for BSP updates which has just run.

kueken:
ok

PR #928, opened 2021-03-05
build optmise for Size (mipsel)

kueken:
Re -Enable all needful Things


kueken:
Sorry, now I accidentally put 2 commits into a pull request again.

Hopefully the ffmpeg built in this way is a good compromise for mipsel.
Feedback from the users in the forum would be desirable here.
greetings

WanWizard:
It's also a PR against the release branch, not the develop branch :wink:, so I can't merge it.

kueken:
I wasn't sure if you can make changes in the 8.
I close it again.
So I adapt it in the develop for mipsel.
Feedback is also important here, whether everything goes well.
greetings

WanWizard:
The problem with changes in a release image is that backporting to develop is often complex. 

Even if it is urgent, I'd rather push an emergency release through than trying to backport...

kueken:
No problem, didn't know exactly whether it was also possible here.
I'll update to develop in a moment.
The aim is to reduce the ffmpeg size (--enable-smal) for mispel.
The construction of the ffmpeg then takes a little longer and (hopefully) the important functions are retained.
I'm just running the construction.
Then comes the pull request.
greetings

PR #939, opened 2021-03-28
update gst ->1.18.4

kueken:
switch back nettle ->openssl
clenaup build


WanWizard:
What is the reason for the revert?

kueken:
sorry it was mistake i will reset git to gst commit

kueken:
So now again gst update individually and that should work too.
However, again leads to about 3mb more with the libav images.
The images are then about the same size as with the pli 7.3. I could not find any problems with the gst and mipsel box regarding streams. But someone will surely find a mistake.
Maybe you should create this separately as a topic in the forum.
So gst errors / requests and ffmpeg errors / requests.
greetings

PR #941, opened 2021-04-01
switch Source and build uchardet

kueken:


WanWizard:
Any particular reason for this switch?

kueken:
cgit is often not available
and therefore followed Open-Embedded
https://github.com/openembedded/meta-openembedded/tree/master/meta-oe/recipes-support/uchardet

A change to the github source would also be possible, but I think it is easy to follow the source from the open embedded.

WanWizard:
We've noticed, hence the reason the recipe had an override option. I'll add the reason to the commit text.

PR #947, opened 2021-04-08
update python-js2py->0.71

kueken:
update python-lzma ->0.0.14
update python-pysmb ->1.2.6
update python-websocket ->0.58


WanWizard:
I've manually merged them because I needed to split this PR into 4. 

Please keep to one change per PR, so it can be easily reverted if needed :grin: .

kueken:
ok

PR #976, opened 2021-05-04
fix wifi for build

kueken:


WanWizard:
You're deleting quite a lot here, does this still build for all supported kernel versions?

kueken:
That looks like more than it is.
I can't build images for everyone.
I built kernel 4 and kernel 5.12 images for kernel 3.
et9x00
fusionhd
h7
osmio4k +
but it builds through completely.
Hopefully there are no other errors either.
Patches were adapted and recreated, since I assumed the current git in each case.

greetings

kueken:
oh...i see one (PR ="2")
I'll take that out.
So you can react faster ;-)

WanWizard:
In the meantime I'll setup a bitbake compile for these drivers, for all boxes. May take a while :grin:.

kueken:
Of the deleted patches, 3 are no longer necessary due to updates in git.
2 I created together in one patch (new)
hope that it also builds for all receivers
I'm running out of memory on my laptop ;-)

WanWizard:
We'll see what happens here. It may take a while, it starts compiling kernels first...

WanWizard:
Was too quick, can't build them for all boxes, they don't build for kernels < 4.0.

Guess it's best to merge it, and see what happens tonight...

kueken:
Kernel <4 
test with build et9x00 Kernel3.8 is ok

WanWizard:
Does that mean we need to review the versions in https://github.com/OpenPLi/openpli-oe-core/blob/develop/meta-openpli/recipes-openpli/images/openpli-enigma2-feed.bb#L26 ?

kueken:
Don't know exactly what you mean

The feed with the old kernel works without any problems, that's what I wanted to say.

kueken:
gbue4k is currently building, but that takes a while with my ryzon5
build feed ok
h9 start......end
build feed ok

kueken:
offtopic on
meta qt5 could be brought up to date 'zeus'.
For any 8.1

WanWizard:
> Don't know exactly what you mean
````
	${@ 'kernel-module-8192fu' if ("${KERNEL_VERSION}" and bb.utils.vercmp_string("${KERNEL_VERSION}", '4.0') >= 0) else '' } \
	${@ 'kernel-module-8821cu' if ("${KERNEL_VERSION}" and bb.utils.vercmp_string("${KERNEL_VERSION}", '4.0') >= 0) else '' } \
	${@ 'kernel-module-88xxau' if ("${KERNEL_VERSION}" and bb.utils.vercmp_string("${KERNEL_VERSION}", '4.0') >= 0) else '' } \
````

These kernel modules are only build for kernels >= 4.0

kueken:
mhh...
I'll see if I can fix something

kernel-module-8821cu is building with Kernel 3.8 (et9x00)
8192fu is classic problem (extern-inline and old Kernel)

kueken:
8192fu fix for build Kernel <4

PR #979, opened 2021-05-07
distro: add 'mplt' to DISTRO_FEATURES - fix kodi for Vu+ mipsel

andrea-adami:
The toolchain needs to be compiled with --with-mips-plt
Otherwise the vendor-provided drivers badly link during kodi build and
the binary crashes at start.


persianpros:
Just to let you know this will recompile everything ;)

andrea-adami:
I know it is a pain...but it has been somehow overlooked before :)

PR #983, opened 2021-05-11
kodi_18: split out the stb patchset

andrea-adami:
make it possible to read and review


andrea-adami:
Note that there isn't any change, patch is just divided in chunks.

PR #986, opened 2021-05-16
kodi_18: remove default CMake options

andrea-adami:
see /CMakeLists.txt


WanWizard:
What is the reason for this, what does it change?

For example, removing `ENABLE_INTERNAL_FFMPEG` means Kodi will build its own FFMPEG version, instead of the one already present on the box...

andrea-adami:
These lines were present in the OE-A stb-kodi but are pointless because these do not change anything...
See  https://github.com/xbmc/xbmc/blob/Leia/CMakeLists.txt

About FFMPEG there is a patch for it in the OE-A forks. If I get it right, they optimize it for arm and default to internal for mipsel.
It is one of the patches I would discuss before importing...

WanWizard:
Ok.

PR #987, opened 2021-05-19
update and fix python depends

kueken:
Fix warning blindscan, misplslcnscan, autobouquetsmaker, mediainfo, twistedsnmp

Check build on mips and arm, but not for all.
hope all ist fine


kueken:
@WanWizard 
I know it's a very big pull request, but you can revert one at a time if you want.
You can test it locally in advance.
Everything is fine on my mips box.

greetings

WanWizard:
Where does this update to? A newer Yocto release (stable) or HEAD (unstable)?

We want to release 8.1 this week, and after that the buildservers go offline due to a redesign of our infrastructure, so I'd like to avoid anything that can make the release unstable at this moment.

kueken:
No, it is more the selected required python dependencies that we usually change something anyway (bbappend).
There won't be any really major updates for python 2.7 anymore, so I restructured it a bit and merged it with python 2.7.
No more files have been added, just some updated and the bbappend merged.
In addition, some plugins regarding the build warning have been fixed.
Maybe you will have a look at it locally first if this is not done until 8.1.
No problem.

WanWizard:
needed to revert pycurl, newer versions are no longer python2 compatible.

kueken:
Yes, 
You're right, shouldn't be happening.
Sorry me.
The previous version was the last.
I will adjust it.
You have already revised the commit. thanks

PR #100, opened 2016-09-09
tuxbox-common: Add atsc.xml to transponder lists

athoik:
Requires: https://github.com/OpenPLi/OpenPLi-1/pull/1

PR #989, opened 2021-05-22
kodi_18: cope with vuplus-support pkgs renames

andrea-adami:
we have now a single recipe, vuplus-kodi

 changes:
 - simplify egl patch for mipsel
 - remove xbmc-support tarball from SRC_URI
 - for vuplus, only add minimum runtime deps
  (the shlib resolver seems not catching the
   .so libs without version/soname)

 We see in the log do_package of vuduo2:
 NOTE: Couldn't find shared library provider for libEGL.so
 NOTE: Couldn't find shared library provider for libnxpl.so
 NOTE: Couldn't find shared library provider for libGLESv2.so
 NOTE: Couldn't find shared library provider for libxbmc_base.so

 For Vu+ the runtime dependency is then explicited adding the
 vuplus-kodi support package whose lib RDEPENDS on libgles.

 Other STB have similar issues with the drivers, the solution
 would be to explicit the rdeps but for the moment we avoid
 the QA check about [file-rdeps]. GLES is anyway installed.


andrea-adami:
Please merge after meta-vuplus merges PR #100
Thx
A.A.

PR #991, opened 2021-05-23
fix for texturepacker license check

andrea-adami:
copy and paste error, sorry


andrea-adami:
follows another one for jsonschemabuilder...beer today was too strong...

PR #1003, opened 2021-06-06
gst update 1.18.4 last git commit

kueken:


kueken:
What do you think about the possibility of Autorev here?
We are in the stable branches at ffmpeg and gst.
Errors rarely occur there.

WanWizard:
If that is useful and doesn't create issues, I don't have a problem with that. In releases they will be pinned anyway.

Either way, the SRC_REV should not be in the recipe, but in /meta-openpli/conf/reporefs.conf

kueken:
done.

Iffmpeg-old need for what?

WanWizard:
Looks like a left-over from the Zeus update work, see https://github.com/OpenPLi/openpli-oe-core/commit/faf8d32cdddfa8b29e936a8fa65927c15adb0aeb (the commit text). I think it can go.

kueken:
Done

PR #1011, opened 2021-06-17
V3 for musl

andrea-adami:
Apart dvd+rw-tools the patches have been cleaned and sent upstream.
See patch header for the fixed errors and commit log for the changes.

Thanks for reviewing it!
A.A.


andrea-adami:
sorry again the same typo...good is I sent the correct one upstream.
closin, v4 coming.

PR #1018, opened 2021-07-07
fix makefile->meta-abcom

kueken:


WanWizard:
Forgot I renamed it, already fixed in the meantime.

PR #1019, opened 2021-07-08
add zgemma H10 combo(?)/2s/2h

Dima73:


WanWizard:
Sorry, should have warned you, it doesn't work that way, as it is based on MACHINE (at compile time we don't have runtime data from the box).

PR #1027, opened 2021-07-24
tryfix mips

kueken:
Try to solve the inexplicable problems of some 'Mips' boxes related to the glibc used


kueken:
I only know that some of the 'Mips' have problems.
I can't test this though.
It would be good if someone could test an affected box with mostly inexplicable 'freeze'.
In principle, however, the patch should not have a negative effect. It has already been integrated into the more recent glibc

PR #1029, opened 2021-08-03
Hardknott-update more for building

kueken:
not ready yet


andrea-adami:
kueken, thanks!
I could not yet start helping, maybe in the next days.

I have a couple of questions about directfb for vuplus, forgive my ignorance.

First: why is the recipe in meta-openpli and not in the BSP?
Second: I understand there isn't any accelerated driver for the older vuxxo2/mipsel but only for the arm machines. Correct?

Then, apparently you took from OE-A some code, good, we will clean it later i.e. do_qa_staging() {} and /usr/lib instead of ${libdir}.
Not a big problem at this stage.

Cheers
A.A.

kueken:
Hy

Problems are very different from directfb.
Incidentally, it is much easier to have it in the Meta pli than on the manufacturer's layer, for example.
The biggest problems at the moment are libtalloc and libtdt, i.e. following the Samba.
Samba is officially no longer compatible with python2 (since version 4.10 I think).
Therefore, the environment variables in the current version also cause problems.
I tend to lean on the basically ready-made environment on the part of 'Hains'.
Because of qt and egl, a few things from the OE-Alliance have to be taken into account.
You can't change anything about Kodi because of python2.
Newer Kodi requires Python3.

 this conversion to $ (libdir) was unfortunately ported everywhere by PP and unfortunately makes more mistakes than you might think. Hard code is also better from my point of view.

The question is also whether to fast forward the branch towards the master, or to fix it at hardknott?
@WanWizard 
I also continued to build in the direction of the master, but then you have to patch a lot more directly.
glibc is then immediately a problem again.

Regards

andrea-adami:
About samba, seems 4.10.x in hardknott supports python2/3:

https://git.openembedded.org/meta-openembedded/commit/meta-networking/recipes-connectivity/samba?h=hardknott&id=0c9e727a28cd87025b4208ddadb3213aac0bfdee

kueken:
I am not sure...
But is not there exactly
-> Switch to python3?

Whereby that only concerns the construction.
As of Samba 4.11, python2 is probably end

https://www.samba.org/samba/history/samba-4.11.0.html

Samba3 would always be enough for me ;-)

kueken:
Are we sticking to the 'Hardknott' branch?
To me, this seems much easier and more structured than fixing the master branch somewhere.

andrea-adami:
I mean that for hardknott it seems trivial to revert samba to python2, just look at the last lines of the patch.
Maybe this helps in this first phase.

But you are right, this is a temporary workaround, after hardknott we'll be here again.

WanWizard:
@kueken what exactly was the question? Laptop crashed yesterday, still busy filling in the blanks...

kueken:
Whether we should fix the branch 'Hardknott' according to the OE core, or choose a more up-to-date point following the master branch?

PR #1032, opened 2021-08-10
tryfix kodi

kueken:
multibox(se)
mutant hd2400

just Test for https://forums.openpli.org/topic/84902-kodi-on-mutant-hd2400-possible/


andrea-adami (commented):
Hello,

unfortunately it is not that easy to backport. This patch is not applicable.
In this .inc we exclusively add machine-specific patches and fix the (r)dependencies.
So we do not add any PROVIDES or RPROVIDES here.

The EXTRA_OECMAKE is used differently here   -DWITH_V3D=  and is used in the  egl-1-v3d-mali.patch   

Finally, pls check about  SRC_URI_remove_append_  : one or the other :)

please send a v2, thanks
A.A.

kueken:
I'm confused...
Feel free to adjust that if you know better.
But the comment doesn't help me or the user.
Hisi and Kodi are handled differently due to different drivers.
Hence, they are all equally applied (as in your structure) wrong.
There is not only VU + that should be noted by me.

andrea-adami:
nothing to do with vuplus, is just not applicable.
Cheers
A.A.

PR #1041, opened 2021-08-28
Update ABM

AbuBaniaz:


AbuBaniaz:
Please also cherry pick to 8.0 and earlier versions. Pity you wont allow autorev for this plugin

WanWizard:
Why? They won't be build anymore anyway?

AbuBaniaz:
If the older images are not being built, please ignore that part of request. Would be handy to know which image are actually built.

WanWizard:
Under normal circumstances:
* The latest minor release of the current major ( = 8.1 at the moment )
* The latest minor release of the previous major ( = 7.3 )

But we don't build 7.3 anymore because we've lost the PR database in a hardware crash two months ago.

PR #1049, opened 2021-09-15
update gst & ffmpeg build and Version

kueken:


WanWizard:
Getting
````
ERROR: gstreamer1.0-plugins-good-1.18.5+gitAUTOINC+7d9cdda764-r0 do_patch: Command Error: 'quilt --quiltrc /openpli/oe/release-8.1/build/tmp/work/armv7ahf-neon-oe-linux-gnueabi/gstreamer1.0-plugins-good/1.18.5+gitAUTOINC+7d9cdda764-r0/recipe-sysroot-native/etc/quiltrc push' exited with 0  Output:
Applying patch 0001-Revert-souphttpsrc-Always-use-the-content-decoder-bu.patch
patching file ext/soup/gstsouphttpsrc.c
Hunk #1 FAILED at 1509.
1 out of 1 hunk FAILED -- rejects in file ext/soup/gstsouphttpsrc.c
Patch 0001-Revert-souphttpsrc-Always-use-the-content-decoder-bu.patch can be reverse-applied
```` 
on the current build.

WanWizard:
> What exactly does this message mean?
Quilt is basically trying to tell us that ALL the changes in the patch (we are trying to apply) are already present in our code.The term "reverse-applied" seems to be derived from some other version-control system.

So, the patch can be removed?

WanWizard:
Sorry, ignore this, caused by a rebase issue.

kueken:
ok

PR #1050, opened 2021-09-15
add glibc patches fix trouble with Kodi

kueken:
thx OE-Alliance


WanWizard:
Another full rebuild...

kueken:
Not quite, but 'only' for arm (hisi / BCM) but not for mips.
But it's about patching, I would also transfer it to 8.1.
But do not know how it is wanted.

WanWizard:
Lets see how it builds in develop first (and if it actually works :grin:).

kueken:
functions
Otherwise I wouldn't have sent it as PR ;-)
ask with User doglover;-)

kueken:
Another problem is missing alsa support in many bsp

PR #1056, opened 2021-09-21
Fix Kodi build with use old Version crossguid

kueken:


andrea-adami:
Interesting.
whith this issue I can probably lure Khem and have soon kodi in OE.
Cheers
A.A.

kueken:
crossguid ->old use 
`	install -D -m 0644 ${S}/guid.h ${D}${includedir}/guid.h
`
Kodi18.9 need this for cmake
crossguid "new" use
`	install -D -m 0644 ${S}/Guid.hpp ${D}${includedir}/Guid.hpp`
configure erro stop build

andrea-adami:
As you know my idea is to add meta-kodi and only have a kodi_git.bbappend in OpenPLi.

For the specific issue there is already a patch:
https://github.com/koenkooi/meta-kodi/commit/0e2fa8b9efc908b953ddb088b62958d8fff41f25

kueken:
I really don't understand what you mean?
Version of crossguid is completely irrelevant.
Now where is your problem with this fix?
It's not worth discussing at all, is it?
This is simply a bug on the part of Kodi 18 and so can be easily fixed.

andrea-adami:
My point is that we should add a patch to kodi and not add an older version of crossguid.

andrea-adami:
Please let me finish the round-up of kodi_18 on PLi 8.1 then I'll better understand.
Forgive me but I don't yet understand the plans for python3/kodi_19.

I see the some recipes like libcec are using python3 so I really don't understand if you maintainers don't want to have two versions or what.

Cheers
A.A.

kueken:
It doesn't matter which Crossguid is there.
Kodi uses this internally / externally for c.
As described above, Crossguid has only changed its ending, so that doesn't work with Kodi.
Has been (badly) adapted in Kodi 19. Again, only the previous version works.
But you can easily patch it there.
It doesn't make sense here, you just take the version that works.
All other programs can handle this as well.

This has nothing to do with Python.
Whether programs use python 3 to build has nothing to do with their functionality in python3 or Python2.
There are some programs built with python3, check out Samba.

With Kodi, however, the functionality for Python2 was stopped from version 19. The code was simply deactivated.

andrea-adami:
I am even more confused :)

As for samba, can be built / can run with python2 only and was upgraded to python3 here:
https://git.openembedded.org/meta-openembedded/commit/meta-networking/recipes-connectivity/samba?id=0c9e727a28cd87025b4208ddadb3213aac0bfdee

I was under the impression that python3 was taboo, being enigma2 uses only older p.
So for PLi-hardknott we'll have both python2 and python3, right?

sorry for the noise and thanks!
A.A.

kueken:
Python2 is used in the image.
Therefore we can only use things that support python 2.
Samba uses, among other things, python3 to be built.
It can still be used on python2 and python 3.
At Kodi, this was discontinued from version 19.
Therefore this version cannot be used for the current e2.

I think your problem lies in the differentiation between building a program and using it afterwards.
Your PC can of course build many applications with python3, but these are then intended for python2.

WanWizard:
It would be good to make this known, as there still seem to be some reluctance amongst developers to migrate to Py3.

If OE is moving to Py3, and packages becoming dependent on Py3, it would be silly to keep Py2 around...

kueken:
Whether e2 python3 or python2 has nothing to do with this PR.
Whether you have to switch to Python3 because of Kodi, I'll leave it open.
The actual system is not Kodi.
Kodi still works quite well in version 18

WanWizard:
> Whether e2 python3 or python2 has nothing to do with this PR.

I know that, it was a general observation. The entire world is migrating to Py3 because of the EOL of Py2, it would be silly of us to still maintain part of the code on Py2, if only because of the flash space wasted by both the Py2 and Py3 install.

kueken:
That means, stop working here and wait for enigma2 in python3.
OK

WanWizard:
NOOOO.....

I meant the other way around: make the devs work on E2 in Py3 to be ready for OpenPLi 9 (on hardknott).

andrea-adami:
My final words about this patch after much Ot (my fault)

>My point is that we should add a patch to kodi and not add an older version of crossguid.

kueken:
Is possible, too.
You would have to patch Cmake accordingly.
Actually 'only' the ending has to be changed.
So instead of "guid.h" to "Guid.hpp" .
But that's really only cosmetic.

andrea-adami:
once we upgrade to kodi 19 this will be the single upstream patch

https://github.com/koenkooi/meta-kodi/tree/hardknott/recipes-multimedia/kodi/kodi

then we'll have to add the stb support.

PR #1064, opened 2021-09-30
kodi_18: temporarily add MACHINE_FEATURES in kodi_18.inc

andrea-adami:


andrea-adami:
I have added the renaming to the fix.

Now my doubts are about nxpl, in OE-A this can match with v3d-cortexa15 and v3d-mipsel.
I must check if we really need the split in two.

andrea-adami:
same for nxclient in meta-gi and meta-qviart: I cannot easily find OE-A matches, maybe

dags-cortexa15 or nextv-cortexa15

andrea-adami:
please do build-test with this patches, thanks

andrea-adami:
with this PR kodi builds again  for vuduo4k

PR #1070, opened 2021-10-04
kodi_18: add INSANE_SKIP_${PN}_dual += "file-rdeps"

andrea-adami:
The newly added meta-qviart duo unfortunately arises the issue:

QA Issue: /usr/lib/kodi/kodi-stb contained in package kodi
requires libmali.so, but no providers found in RDEPENDS_kodi? [file-rdeps]

this fo libmali.so, libEGL.so, libGLESv2.so
(as for other BSP's provided drivers)


WanWizard:
Isn't this something that should be adressed in the BSP?

andrea-adami:
yes, but probably the libs lack soname or version so we must add RDEPENDS like for vuplus.

I am piling the affected devices at the end of kodi_18.inc so we'll see which set of drivers are used.

PR #1073, opened 2021-10-11
update files streamlink

kueken:


kueken:
please also take over in 8.1 and then my recommendation there Streamlink pin to the release version

WanWizard:
so pin in develop too?

kueken:
Better to Pin "Billy's" release.
The updates or takeovers of Billy from the original streamlink(python3) branch in the meantime lead to errors again and again.
It is better to follow the release and always update the additional files.

WanWizard:
I have no idea what that means... :grin:

kueken:
https://github.com/Billy2011/streamlink-27
is the python 2.7 fork.
Commits are merged and than adapt to python2.

WanWizard:
Ok. So just to confirm, pin both develop and release on https://github.com/Billy2011/streamlink-27/commit/39f3855b59c2ef01f3d6513a1cc972968494a35c ?

kueken:
In order to avoid mistakes, I would suggest this.
The supplementary 'files' are updated in parallel to the release.

PR #1076, opened 2021-10-18
update OpenEmb-Core

kueken:


kueken:
to build vu + with webkit you would have to switch locally to my fork branch next.

WanWizard:
How come are there BSP files in this PR?

kueken:
Which files do you mean?

At vu + I then have to create a PR. 
I want to wait, because otherwise every develop and 8.1 will be completely rebuilt at this point.
Rebuilding the webkit takes longer than anything else put together. 
Therefore my recommendation for possible tests is to do this locally in advance from the branch.
switch locally for Test
https://github.com/kueken/meta-vuplus/tree/next

a rebase has not been possible for a long time, as some Kodi updates have messed this up a lot.
The BSP has therefore been updated manually.

WanWizard:
There are lots of BSP's in this PR, which changes to conf files for example.  They can't be part of this PR, as BSP's are submodules.

kueken:
That's just the commits, but it doesn't matter.
I close and reopen later without a BSP UPDATE

PR #1077, opened 2021-10-18
update hardknott

kueken:


kueken:
bsp need update

WanWizard:
I've updated all BSP's to the latest version.

PR #1092, opened 2021-12-16
restructure steamlink build and update last release

kueken:
add pycountry
delete is639/iso3166


WanWizard:
You're sure they aren't used elsewhere?

kueken:
Pycountry should be the solution for this.
All iso are included there.
But I check again whether it is still depend somewhere

kueken:
Ok
I can't find any other addiction.
From my point of view, it is no longer needed.
If a plugin specifies this as a Depend, then the pycountry is also possible.

WanWizard:
I asked because the commit message in release-7.3 was
> added missing python packages isodate, iso3166, iso639 and websocket 

which suggests something (a plugin perhaps) needs it.

kueken:
Streamlink needs this ;-)
I have now rebuilt it and added pycountry for it.
At first it was just too big, but with the trick of removing the 'po' files, pycountry is not bigger.

kueken:
;-)

PR #1109, opened 2022-02-23
Add branch to fix transmission fetch

Taapat:
This fix fetcher failure Unable to find revision bb6b5a062ee594dfd4b7a12a6b6e860c43849bfd in branch master even from upstream


WanWizard:
A side effects of the desire of certain groups to remove "master".

PR #1113, opened 2022-06-03
pin python-iso3166 to last Version python2

kueken:


kueken:
Alternatively, switching to pycountry might be a better alternative.

I addressed this a long time ago.
This is also implemented in the Hardknott branch.
for possible cherry pick, I created the commit here
https://github.com/kueken/openpli-oe-core/commit/76cce5fd8d4c022a344e9d303032924d6a72a735

It is also possible that the changes in the livestreamsrv (streamlinksrv) cause problems.
https://github.com/athoik/livestreamersrv/commits/master
Pin is the safe choice for python2 there too.
Pin to
https://github.com/athoik/livestreamersrv/commit/31b440ff1239e7505e9348cf507b09867bdd1b56

WanWizard:
As I understand it is only a change in _init_.py. 

I asked in the forum to post that change (which someone had made) so we could patch streamsrv in the recipe, but I got no response.

Switching to pycountry is probably more "future-proof" than pinning iso3166.

kueken:
ok
 fix Streamlink with new pull request

PR #1123, opened 2022-08-20
Hardknott py3

OpenDelta-Team:
Hi it is possible to compile with python 3 thanks to all the PLi team


WanWizard:
Merging py3 into develop? Don't think so....

WanWizard:
What exactly was it you wanted to do?

PR #1131, opened 2022-10-12
qtwebkit

zgemma-star:
Submodule change back to github.com/meta-qt5
Fixes for boxes without opengl support

Test and works on H8, H9se and H7
New qtstalker, qthbbtv, youtube-tv based on qtwebkit will follow soon.


WanWizard:
With this PR, the compile of qtwebengine fails for several boxes with a HiSilicon chipzet, including your H9.

zgemma-star:
The patch 0002-Replace-hbbtv-responses-with-application-xhtml-xml.patch fails for qtwebengine.
I will check it out why this fails and fix this patch.

zgemma-star:
fix hbbtv patch, now webengine build as well ok

kueken:
Thanks for the customization

I've adopted a few more things, one patch is currently disabled because I'm not sure if the patch will cause more problems.

Had been preparing the update locally for some time, but it was a little more difficult because of the problems with the different BSP layers.

PR #1134, opened 2022-10-19
Enable libxml2 for ffmpeg

skyjet18:
Enable support for libxml2. It is needed for mpd dash demuxer to work.


skyjet18:
In actual builds there is mpd demuxer enabled, but it is not working without libxml2 support, because without it mpd manifest parser doesn't work at all. This change enables support for mpd dash in ffmpeg and also in exteplayer3. It would be great to merge this change also to actual stable branch.

PR #1153, opened 2023-01-26
update qt pin

kueken:


WanWizard:
Any particular reason this must be updated in a release image? Anything broken now, that this fixes?

kueken:
https://github.com/zgemma-star/e2plugins/commit/794a44d80155dd435d55d876e8bd4acc40935377

Whether it is necessary?
I dont know.
I followed the update, the fix.
It's only the plugins that get updated, not the QT itself has changed.

PR #1165, opened 2023-04-11
Optimize SSD mount in udev

Taapat:
Set deadline scheduler for non-rotating disks.
Use the noatime mount option for SSD.

### DIFF
--- a/meta-openpli/recipes-core/udev/eudev_%.bbappend
+++ b/meta-openpli/recipes-core/udev/eudev_%.bbappend
@@ -2,7 +2,12 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
 
 SRC_URI += " \
 	file://init \
+	file://60-ssd-scheduler.rules \
 	"
+
+do_install_append() {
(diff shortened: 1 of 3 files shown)


neo-pli-bsps:
"The noatime option is no longer considered a must-have for SSDs. On Linux, each file operation updates the access time (atime) of the file. The noatime mount option omits this step, thus saving write ops. But some programs rely on an atime update and do not work with noatime. Two new mount options therefore combine performance with functionality and make noatime superfluous: relatime became the default setting in kernel 2.6.30. The relatime option tells Linux to only update the atime if the file has changed. lazytime is a new option in kernel 4.0. lazytime keeps updates in RAM that are flushed later along with other changes."

Perhaps relatime is better (lazytime is no option with the old kernels of some boxes)?

Taapat:
Do not forget that the hdd is not used for the root system as usual in linux, but mostly only as media storage.
We already mount network drives with noatime without any negative effects.
Therefore, I don't see a problem to also mount a local ssd drive with the noatime option.

PR #1198, opened 2023-08-17
Update reporefs.conf ->enigma2-plugins->lcd4linux

kueken:


WanWizard:
Done because of the easymedia fix which supersedes this one.

PR #1216, opened 2023-11-15
sabnzbd: Update bash script, thx hvrshvrs

Hains:
https://forums.openpli.org/topic/96430-openpli-90-release-beschikbaar/page-7#entry1579422


WanWizard:
Does this need backporting to 9.0-release ?

PR #1225, opened 2024-01-02
[Added] enigma2-boxlogos recipe

DimitarCC:


WanWizard:
I have to close this, already working on it, as we need to deal with box related data at runtime, not build time.

PR #1228, opened 2024-02-05
wireguard: fix depends and fix wg-quick

athoik:
This commit fixes lot of report issues with wireguard.

Issue 1: /usr/bin/wg-quick: line 32: resolvconf: command not found Fixed by adding an RDEPEND to openresolv

Issue 2: AllowedIps = 0.0.0.0/0 fails with ip: invalid argument '51820' to 'table' Fixed by adding an RDEPEND to iproute2-ip

Issue 3: /usr/bin/wg-quick: line 32: ip6tables-restore: command not found Fixed by adding iptables existance check in wg-quick

Issue 4: - nothing provides wireguard-module needed by wireguard-tools Fixed by depending only on kernel-module-wireguard. Now wireguard-module RPROVIDES kernel-module-wireguard for kernels < 5.6

### DIFF
--- a/meta-openpli/recipes-kernel/wireguard/files/wg-quick.patch
+++ b/meta-openpli/recipes-kernel/wireguard/files/wg-quick.patch
@@ -1,0 +1,13 @@
+diff --git a/src/wg-quick/linux.bash b/src/wg-quick/linux.bash
+index 4193ce5..0d85840 100755
+--- a/src/wg-quick/linux.bash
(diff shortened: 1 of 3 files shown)


athoik:
@WanWizard  please note that you already fixed topic 4 by doing a conditional depends.

Mine is doing a conditional runtime provides.

Same fix, different way of thinking...

PS. No need fo PACKAGE_ARCH = "${MACHINE_ARCH}" as well ...

PR #1236, opened 2024-03-22
libdvbcsa: more precise optimization for cpus

WXbet:


WanWizard:
Not sure these are (all) relevant, most of them are X86 ?

WXbet:
You can strip it down.

WXbet:
@WanWizard:

image
This option is deprecated since [11754](https://trac.streamboard.tv/oscam/changeset/11754) and will prevent oscam from startup successfully.

```
unrecognized option '--utf8'
ERROR: Unknown command line parameter: --utf8
```

PR #1250, opened 2024-04-29
Change plugin youtube pinned version to AUTOREV

Taapat:
Recently, to work correctly, an plugin update is needed almost every day.
Therefore, I don't see any reason to pin it and then update it regularly.


WanWizard:
As long as you're sure you're not committing anything that will break the build...

PR #1276, opened 2025-01-29
Force update MACHINE h9twin(se)/more ET(Xtrend)

Dima73:
more info:
https://forums.openpli.org/topic/101067-zgemma-h9-twin-se-enigmainfo/

https://forums.openpli.org/topic/101409-pli-91-on-et4000-installation-failure/?view=findpost&p=1668085

### DIFF
--- a/meta-openpli/recipes-openpli/enigma2/enigma-info.bb
+++ b/meta-openpli/recipes-openpli/enigma2/enigma-info.bb
@@ -479,6 +479,7 @@ elif [ "$MACHINE" = "h9combo" ]; then
 	if [ "$model" = "h9twin"  -o  "$model" = "h9 twin"  ]; then
 		updateinfo "displaymodel" "H9 TWIN"
 		updateinfo "machinebuild" "zgemmah9twin"
+		updateinfo "machine" "h9twin"
 	elif [ "$model" = "h9combo" ]; then
 		updateinfo "displaymodel" "H9 COMBO"
 		updateinfo "machinebuild" "zgemmah9combo"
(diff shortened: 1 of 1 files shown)


WanWizard:
This is not correct and should not be merged,

MACHINE is the variable used to build, and should not be altered. The problem is old code in Enigma.

Dima73:
Not change MACHINE
update only printf "machine='${MACHINE}'\n" >> ${D}${INFOFILE}

WanWizard:
That is changing the MACHINE value.

I should have been clear, what should change is the "model=" line, and Enigma should use that instead of machine, which is a mistake @littlesat made when he make de first enigma.info implementation.

littlesat:
There are too many confusing
Names in enigma.info.... that Triggers mistakes

Dima73:
Need use machine
not need use model

example:

enigma.info
machine='gbquad4k'
model=gb7252

 


enigma.info.bb
MACHINE_MODEL ?= "${MACHINE}"

printf "machine='${MACHINE}'\n" >> ${D}${INFOFILE}

printf "model=${MACHINE_MODEL}\n" >> ${D}${INFOFILE}

WanWizard:
No. 

machine contains the OE MACHINE name which was built, and should not be altered in any circumstance.
if model does not contain the correct value, it needs to be addressed in enigma-info.bb. 

Many zgemma's are a prime example, the machine is "h9", the model can be "h9s", "h9t", or "h92h".

It needs the entire old "if model ==a then model = b" tree that was in Enigma (or is in OWIF) implemented because unlike OE-A images, we don't have a one-on-one between machine and model. I never received any support from the python side when I was implementing this, so it never happened.

THAT is what needs to be fixed here.

kueken:
https://github.com/oe-alliance/oe-alliance-core/commit/2dd9f0955fe61aa74102a7b2088b04d2f2caee7e

Could this be a solution?

WanWizard:
OE-A needs that solution because all other info comes from the static configuration in MACHINE conf files. 

We can't do that (in that way), as we don't have a one-on-one between MACHINE and model. For us, code like this has to go into enigma-info.bb.  A lot of it already is, but it is not complete as my plate was (and still is) overloaded and I didn't get any help. Ideally, someone should take like the OWIF owi model code (that long long list of "if model = then"), and convert all changes made into enigma-info data (either in the buildtime or runtime section).

I also think model and displaymodel should contain the same, only model is a unified value and displaymodel is a formatted string for display purposes. So if displaymodel=ET4000, then model should be set to et4000, and not left at et4x00.

PR #1278, opened 2025-01-31
Rename  "model" as machinemodel

Dima73:
-rename  "model" as "machinemodel" (machinemodel='${MACHINE_MODEL}'\n" >> ${D}${INFOFILE}) 
-add simple variable "model" (model='${MACHINE}'\n" >> ${D}${INFOFILE}) 
-update "model" for  h9twin(se)/more ET(Xtrend)


littlesat:
Why do you need to name to machinemodel? That is also a parameter that does not exist in the enigma.info specs. When you do not do this I can merge.
Could also be I'm confused as by brain also starts to explode

WanWizard:
I can't remember where MACHINE_MODEL comes from. I can see I added it "to sync with OE-A", but if I look at the current OE-A code, it doesn't define it at all.

So I think it should be removed completely (including the original assignment and the printf line) to avoid further confusion.

littlesat:
I could use it for downloading images.... but we have already something at github that helps to converts the machine/model names for download foreign images.

PR #1290, opened 2025-03-04
Zgemmma update bsp

kueken:
fix h11 missing blindscan-dvbs and transcoding


WanWizard:
Don't bother with 9.1, soon we''ll release 9.2 from develop, and all this work is for nothing.

kueken:
9.1 is wrong or incorrect in some areas.
Especially in connection with the Machine_Fetaures.
Why shouldn't you repair this error there?

I don't understand your argument.
That should help users and yet makes sense to repair it there.

Apart from that, was it intended that you just not create another 9 release?
Why more release, although they are only necessary repairs of a faulty release?

The branch itself is no longer changed in terms of OE-Core.

I would urgently suggest this to update all of this in 9.1 so as not to make it difficult for users to make it unnecessarily difficult.

WanWizard:
Because as soon as we promote develop to 9.2, 9.1 will seize to exist. But it is your time...

PR #1307, opened 2025-05-08
fix Syntax

kueken:
PACKAGE_ARCH := -> PACKAGE_ARCH =


WanWizard:
Maybe double check, as there is a difference:

= means expand variable when  actually used
:= means expand immidiately (when parsing).

In most causes, there is no functional difference, but there could be, if something else is manipulating the variable or its parent value.

But I think this is ok, MACHINE_ARCH is a pretty stable value.

PR #1316, opened 2025-08-02
Update OE-A version of bb files for xstreamity and xclass

AbuBaniaz:
Use git tags to avoid version issues

### DIFF
--- a/meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-xklass.bb
+++ b/meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-xklass.bb
@@ -1,24 +1,24 @@
 SUMMARY = "IPTV Xtream Codes/XUI One playlists player by KiddaC"
 HOMEPAGE = "https://www.linuxsat-support.com"
 MAINTAINER = "kiddac"
+PRIORITY = "optional"
 require conf/license/license-gplv2.inc
+require conf/python/python3-compileall.inc
 
-RDEPENDS_${PN} = "${PYTHON_PN} ${PYTHON_PN}-multiprocessing ${PYTHON_PN}-requests \
-                  ${PYTHON_PN}-image ${PYTHON_PN}-imaging ${PYTHON_PN}-pillow wget \
-"
+RDEPENDS:${PN} = "${PYTHON_PN}-pillow ${PYTHON_PN}-requests wget"
 
-inherit gitpkgv allarch python3-compileall
+inherit gittag allarch
 
 SRCREV = "${AUTOREV}"
+PV = "git"
+PKGV = "${GITPKGVTAG}"
+PR = "r1"
 
-PV = "git${SRCPV}"
-PKGV = "${GITPKGV}"
-
-SRC_URI = "git://github.com/kiddac/XKlass.git;protocol=https"
+SRC_URI = "git://github.com/kiddac/XKlass.git;branch=master;protocol=https"
 
 S = "${WORKDIR}/git"
 
-FILES_${PN} = " ${libdir}/enigma2/python/Components/Converter/* \
+FILES:${PN} = " ${libdir}/enigma2/python/Components/Converter/* \
                 ${libdir}/enigma2/python/Components/Renderer/* \
                 ${libdir}/enigma2/python/Plugins/Extensions/XKlass/*"
 
@@ -36,4 +36,3 @@ cp -rf ${S}/XKlass/usr/lib/enigma2/python/Components/Converter/*.py ${D}${libdir
 cp -rf ${S}/XKlass/usr/lib/enigma2/python/Components/Renderer/*.py ${D}${libdir}/enigma2/python/Components/Renderer/
 cp -rf ${S}/XKlass/usr/lib/enigma2/python/Plugins/Extensions/XKlass/* ${D}${libdir}/enigma2/python/Plugins/Extensions/XKlass/
 }
-
--- a/meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-xstreamity.bb
+++ b/meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-xstreamity.bb
@@ -1,24 +1,24 @@
 SUMMARY = "IPTV Xtream Codes playlists player by KiddaC"
 HOMEPAGE = "https://www.linuxsat-support.com"
 MAINTAINER = "kiddac"
+PRIORITY = "optional"
 require conf/license/license-gplv2.inc
+require conf/python/python3-compileall.inc
 
-RDEPENDS_${PN} = "${PYTHON_PN} ${PYTHON_PN}-multiprocessing ${PYTHON_PN}-requests \
-                  ${PYTHON_PN}-imaging ${PYTHON_PN}-pillow \
-"
+RDEPENDS:${PN} = "${PYTHON_PN}-pillow ${PYTHON_PN}-multiprocessing ${PYTHON_PN}-requests"
 
-inherit gitpkgv allarch python3-compileall
+inherit gittag allarch
 
 SRCREV = "${AUTOREV}"
+PV = "git"
+PKGV = "${GITPKGVTAG}"
+PR = "r1"
 
-PV = "git${SRCPV}"
-PKGV = "${GITPKGV}"
-
-SRC_URI = "git://github.com/kiddac/XStreamity.git;protocol=https"
+SRC_URI = "git://github.com/kiddac/XStreamity.git;branch=master;protocol=https"
 
 S = "${WORKDIR}/git"
 
-FILES_${PN} = " ${libdir}/enigma2/python/Components/Converter/* \
+FILES:${PN} = " ${libdir}/enigma2/python/Components/Converter/* \
                 ${libdir}/enigma2/python/Components/Renderer/* \
                 ${libdir}/enigma2/python/Plugins/Extensions/XStreamity/*"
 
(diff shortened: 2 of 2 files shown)


AbuBaniaz:
Please double check before merging

WanWizard:
I am against changes like this, simply to accomoidate users downloading and installing stuff manually that is also in the feed. Typical OE-A style work-around.

If the feed is out of date, the user can report it, and we can update it.

Besides that, were does it end, there are thousands of packages in the feed, all have different version numbers from packages build by others. OE-A doesn't even use a PR server but manually defined PR's.

AbuBaniaz:
But that risks breaking things and means there are constant pin update requests.

This was a proper fix so version numbers issue is resolved.

WanWizard:
That's the way it works. If they want to use bleeding edge, they can install a develop image, which has nothing pinned. 

And it was a hack for two specific packages, not addressing all other packages with the same issue. 
I don't see a need the need for it, why should we support packages build by someone else, while we provide them in the feed?

WanWizard:
You need to get your head out of the OE-A "work-around" mindset. The problem here is:
- User used an OpenPLi release image which is known to have fixed versions for everything
- User uses a specific plugin
- User has a (hopefully) genuine requirement for a newer version of said plugin
- User doesn't report this, but starts working around it
- User runs into an issue with this work around
- You think it is up to use to fix this

AbuBaniaz:
It's not about supporting packages, it's about ensuring that package on feed has correct version number.  I hope you can reconsider it.

As you know, I know very little. But I could manage copying and pasting those two. It's beyond my abilities to amend all the bitbake files and add tags.

WanWizard:
No, that is not what you want. You want OUR build to use the same versioning system as OTHER builds. 
To facilitate a work around for a user that can't be bothered to even report it.

AbuBaniaz:
It's not a workaround. The workaround is to use a dxhold command.

The user couldn't even report issue correctly. He kept on saying a reboot caused the plugin to regress. Then he said updates did it. That's when checked and confirmed that the newer version of plugin is being regressed owing to different numbering/versioning.in use. He didn't even report it on the PLi forum.

How about using the same numbering system as the plugin author? Please compare the numbers.Please forget the other builds and do what you think is correct and proper.

littlesat:
I agree with wanwizard this is not the way to do it...

AbuBaniaz:
Any update on what you propose to do to deal with the version numbers PLI uses? I honestly dont know why you reject the use of git tags. I hope the better solution you have in mind can be implemented soon.

Pli uses version numbers like this:
```enigma2-plugin-extensions-xstreamity_576+6bf7630-r0.1_all.ipk```

The Plugin author and OEA uses versioning like this:
```enigma2-plugin-extensions-xstreamity_5.09.20250813_all.ipk```

WanWizard:
You're not supposed to use OE-A ipk's in OpenPLi, and vica versa (for packages in the feed), so this is a non-issue.

Even if you wanted, you can't get it to match, because OE-A hardcode all versions and revisions, and we use a PR server.

AbuBaniaz:
We are talking about ipks that Kiddac builds and uploads. Even though people install newer ones, the older PLI ones replace them because of the higher version number.

The request is to just add the git tags. But you call that version horny.

AbuBaniaz:
OEA do not hard code versions, it uses git tags.  That is beside the point.

PLI uses git tags for plain Oscam. Oscam emu uses them also although the 1.20 is added. So I don't see why you object to git tags being used.

WanWizard:
I simply don't see the point, as you should NEVER try to install an ipk from another build on an image, no matter what image. Unless you know EXACTLY what you do, in which case the version is irrelevant. 

I don't see why we should put effort in making sure our versioning compatible with others, when there is zero reason to do so? If it is so important to you, have OE-A change their versioining to match ours.

AbuBaniaz:
The ipks in question are the ones kiddac uploads. He does not use the OE-A cuild system, he manually creates the ipks.  After  couple of days of testing he commits the code to Github. You use pinning. And when PLi users install newer ones, a software update downgrades them. The same applies to the develop branch.

If you are not going to be reasonable and going to continue to mis-direct, can you then at least use a version number lower that `5.09.20250813`. The `576` number **hardcoded in PLI** is part of the problem.

Once again, and again, and again, this has nothing to do with OE-A ipks.  OE-A images had the same problem until the git tag solution was implemented.

WanWizard:
So the version number is the problem, not git hashes?

But still, we have those plugins in the feed, so there is never a need for end users to install downloaded ipk's from elsewhere. If the problem is the pinning in release images, the root cause of the problem is not the version, but the fact nobody can be bothered to report and ask for an update.

For any issue, find the root cause, and fix that. And not try to work around it.

AbuBaniaz:
Sorry about the delayed response.

>So the version number is the problem, not git hashes?

Yes, but sort of and needs context.  The bitbake file makes the version number. Nobody is a fortune teller and can predict when each image team will make the commits nor what hash GitHub will assign.

As I have said several times and which I seem to fail to explain in a way you accept, Kiddac does not build his plugins on any image team's build system. It took a long time to convince him to host his skins and plugins on Github. When the  problem with version numbers affected OE-A image teams, it was a challenge to get him to adopt the tags system. 

So the root problem for this issue is:
Bitbake file OpenPLI has creates a conflicting version number. 

By using git tags, the problem will be solved. The following pull request has been open for 3 weeks. Please merge it in develop, if there is a problem, we can revert it. And after some testing and it meets your requirements, hopefully yo will merge it into release. 
https://github.com/OpenPLi/openpli-oe-core/pull/1318

There is nothing wrong with using Git Tags.

WanWizard:
The problem I still have with this all, is that the only reason you want this, is that you want to install an IPK build by team A for image A on image B build by an entirely different OE environment. 

And you can't be bothered to either uninstall what is already installed, or use --force-reinstall (in both cases the version of the package is irrelevant).

AbuBaniaz:
Once again, it's nothing to do with teams.  Kiddac is not part of any team, he builds these on his PC and does  not have a build environment.

Also, as said already more than once, if a user installs a newer version, a software update of PLi regresses the plugin. This is because the version number the bitbake file creates is higher. That is the reason why I asked for this.


So we start the circle again, 
PLi uses hashes to ensure stable version
Coder should push update of repo refs
But you should use autorev
We only have stable

Please have some trust in what I am saying. It will be fine. If there is an issue, I will be among first to ask you to revert.

WanWizard:
Sorry, but am I talking to a brick wall?

If there is a new release available, and that is properly tested, that can be reported so the pinning can be updated.

The user SHOULD NOT install versions of packages in the feed manually, and if the user does, the user should know what the fuck that user is doing. So either remove what is installed first, or use --force-reinstall.

The root cause of the problem here are:
* nobody can be bothered to ask for a version update
* people start fiddling around with downloaded ipk's without having a clue what they are doing

And I do see why we should make changes to accomodate lazy and/or ignorant people.

AbuBaniaz:
The root problem is the version number the bitbake file gives is wrong.

If Kiddac uploads his version on Saturday which user installs. On Monday he updates PLi, the version on feeds will downgrade it.

A hash update in reporefs  just rebuilds the version on git with wrong version number.

Instructions to dxhold have already been posted to work around this version issue. But it is better we fix things properly.

WanWizard:
If you only use plugins from the feed, the version is complete irrelevant.

The version string contains the git commit sequence number, which is always unique and only increments, so there can never be a situation where the feed will contain a version that is lower than that already installed. That can only happen if you install something not build by us.

And again, I don't care about people hacking their image by installing third party ipks. If they don't know what they are doing, that is their problem. 

There is never a need to do this. Use what is offered in the feed: problem solved.

AbuBaniaz:
And the circular, irrelevant, deflective arguments restarts/continues.

They are ipks built by the coder. Id call that 1st party.  Installing answer ipk from the plugin creator is not hacking. 

The wrong version number created by your bitbake file is the problem.

Nevermind.

WanWizard:
There is that brick wall again.

The version number isn't your problem,. the fact that someone wants to install foreign build ipks for packages in the feed is your problem. 

And my response is that it is a problem of your own making. Don't, there is no need, problem solved.

kueken:
> If you only use plugins from the feed, the version is complete irrelevant.
> 
> The version string contains the git commit sequence number, which is always unique and only increments, so there can never be a situation where the feed will contain a version that is lower than that already installed. That can only happen if you install something not build by us.
> 
> And again, I don't care about people hacking their image by installing third party ipks. If they don't know what they are doing, that is their problem.
> 
> There is never a need to do this. Use what is offered in the feed: problem solved.

I had already discussed it in the Scartghap branch when committing and I don"t think I need to do it again now.

The PR makes sense in my eyes.

kueken:
https://github.com/OpenPLi/openpli-oe-core/pull/1318

AbuBaniaz:
You are being the brick wall wanwizard.This is so uncharacteristic of you. You tend to advocate for correct and proper implementation. Using Git tags is correct.

Correcting the version number generated by the bitbake file will ensure that those who are at your mercy can remain as they are. Those who decide to install a newer version from the developer can do so without having to mess around with force/override/downgrade command line options.

A software update should never result in a downgrade of packages.

Anyway, I do hope you can at least try the change in the other pull request at least on your.build environment.

littlesat:
The issue here is that this is a work-a-round when you upgrade from a third party source/feed... this was never intended. So when you indeed hack or work-a-round how it should work then indeed you need to 'block' it. It is what it is. As far I read this discussion this looks indeed as a good idea... but this is something we should not consider. When we want the latest version and there is a good reason for it then it is better to request for re-pinning it in a release then everyone gets the latest version after an upgrade/update. So basically no need for this work-a-round. And please please please 'break down the wall!'

kueken:
I really don"t know why you"re arguing here.
The commit was simply wrong and therefore not suitable.
He would have caused construction errors.

I"ve created a corrected PR that allows versioning via gittag.
Applying this, regardless of whether you feel it is necessary or not, is not worth discussing in my view.

Please don"t argue anymore

PR #1318, opened 2025-08-26
Kiddac plugins switch use gittag

kueken:


kueken:
@AbuBaniaz 

I suspect that this was your concern.
Versioning via gittag is also possible in pli and has been done for a long time.

After application, the "tag" is also displayed as the first thing at ipk.

AbuBaniaz:
Thanks, much appreciated. Hopefully it is accepted.

I noticed one of the BB files in your pull request  has autorev removed. Probably by accident on the Developer branches.

kueken:
https://github.com/OpenPLi/openpli-oe-core/blob/develop/meta-openpli/conf/distro/reporefs.conf#L220
You can find the corresponding autorev here, that was already there, so I cleaned up directly to it

PR #1363, opened 2026-05-14
Zgemma BSP Update

kueken:
add h17 Twin


WanWizard:
Sorry, BSP update done manually before I saw this PR. Thanks for the BSP update.

PR #1374, opened 2026-07-24
liquid-dsp - disable SIMD when NEON is not in TUNE_FEATURES

ims21:
liquid-dsp assumes that NEON is available on all ARMv7 targets and enables BUILD_NEON even when the selected tune only uses VFP. This causes compilation to fail with "target specific option mismatch".

Pass --enable-simdoverride for targets without NEON in TUNE_FEATURES so that the portable implementation is used.


ims21:
Work around target specific option mismatch when building for ARM targets without NEON by enabling --enable-simdoverride only when neon is not present in TUNE_FEATURES.

ims21:
So, among the receivers I build for, the problem only occurs on the sfx6008. Better will be fix this box only...

PR #1379, opened 2026-08-05
Update hash for xstreamity and ABM

AbuBaniaz:
5.58-20260720


littlesat:
Why was this mandatory for Release?

WanWizard:
It is not uncommon to bump plugin versions in a release image, after testing.

Both are quite popular plugins, and both had significant updates and fixes.

PR #1381, opened 2026-08-21
[openpli] add GT IPTV Player Pro v1.0.0

VicTuS59:
## Summary

* add a source-built BitBake recipe for GT IPTV Player Pro v1.0.0
* add `enigma2-plugin-extensions-gtiptvplayerpro` to the optional OpenPLi Enigma2 feed packages
* pin the recipe to the stable `v1.0.0` source commit for reproducible builds

## Plugin

GT IPTV Player Pro is a standalone, account-oriented IPTV browser and player for Enigma2. It supports user-supplied Xtream Codes/M3U accounts and Stalker/MAC portals, with dedicated Live TV, Movies and Series screens.

The repository does not bundle channel lists, provider accounts, credentials or copyrighted media. The plugin does not contain a self-update mechanism; feed/package-manager updates remain authoritative.

Source: https://github.com/VicTuS59/GT-IPTV-Player-Pro

Release: https://github.com/VicTuS59/GT-IPTV-Player-Pro/releases/tag/v1.0.0

License: GPL-2.0-or-later

## Packaging

* package architecture: `all`
* installs Python source under `Plugins/Extensions/GTIPTVPlayerPro`
* uses OpenPLi's `python3-compileall` class
* declares the required split Python runtime packages
* removes residual plugin runtime/bytecode files on uninstall without deleting user playlist data

## Validation

* `v1.0.0` resolves to commit `8ce3454e293247263879f34b22df3144a8a0ecb3`
* the source license checksum was verified
* source installation and Python bytecode compilation were reproduced from the tagged checkout
* Python 3.9 grammar compatibility and Python 3.13/3.14 removed-API compatibility were checked
* manual installation, operation and uninstall testing completed on OpenPLi Enigma2 receivers

### DIFF
--- a/meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-gtiptvplayerpro.bb
+++ b/meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-gtiptvplayerpro.bb
@@ -1,0 +1,53 @@
+SUMMARY = "Standalone IPTV browser and player for Enigma2"
+DESCRIPTION = "GT IPTV Player Pro provides Xtream Codes, M3U and Stalker/MAC portal browsing and playback."
+HOMEPAGE = "https://github.com/VicTuS59/GT-IPTV-Player-Pro"
+MAINTAINER = "VicTuS59"
+LICENSE = "GPL-2.0-or-later"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=b234ee4d69f5fce4486a80fdaf4a4263"
+PRIORITY = "optional"
+
+inherit allarch python3-compileall
+
+SRC_URI = "git://github.com/VicTuS59/GT-IPTV-Player-Pro.git;protocol=https;branch=main"
+SRCREV = "8ce3454e293247263879f34b22df3144a8a0ecb3"
+
+PV = "1.0.0"
+PR = "r0"
+
+S = "${WORKDIR}/git"
+
+RDEPENDS:${PN} = "\
+    python3-core \
+    python3-crypt \
+    python3-datetime \
+    python3-difflib \
+    python3-html \
+    python3-io \
+    python3-json \
+    python3-netclient \
+    python3-stringold \
+    python3-threading \
+"
+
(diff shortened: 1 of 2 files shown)


meta-openpli/recipes-openpli/enigma2-plugins/enigma2-plugin-extensions-gtiptvplayerpro.bb
```diff
@@ -0,0 +1,53 @@
...
+HOMEPAGE = "https://github.com/VicTuS59/GT-IPTV-Player-Pro"
+MAINTAINER = "VicTuS59"
+LICENSE = "GPL-2.0-or-later"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=b234ee4d69f5fce4486a80fdaf4a4263"
+PRIORITY = "optional"
+
+inherit allarch python3-compileall
+
+SRC_URI = "git://github.com/VicTuS59/GT-IPTV-Player-Pro.git;protocol=https;branch=main"
+SRCREV = "8ce3454e293247263879f34b22df3144a8a0ecb3"
```
WanWizard, line 12: Is there a reason to pin the build on a SRCREV?

In general, we build plugins from HEAD in development, so issues can be spotted quickly, even though that means that sometimes we build WIP if developers commit before testing.

For release images, we do pin all plugins on a SRCREV, but we use a separate file for it.

Ok for us to remove this?

Also, is this plugin compatible with Python 3.9? If so we can backport this recipe to OpenPLI 9.

VicTuS59: Yes , you may remove the SRCREV line. The plugin is compatible with Python 3.9.

WanWizard: Thanks. I've passed it on for review.

WanWizard:
We've look at it, but is seems AI generated. Nothing is standard, translations are very bad, and using json files instead of gettext, except xstream nothing much works as it should, so in this state we will not add it.

PR #1382, opened 2026-08-22
[meson] fix python install path

ims21:
Fix wrong Meson 0.59.2 Python install path which creates /usr/usr.

The original OE-core patch does not apply cleanly to Meson 0.59.2, so it was ported.

Tested with python3-pycairo:
- /usr/usr is no longer created
- files are installed to /usr/lib/python3.9/site-packages
- do_package succeeds
- python3-pycairo package is created

Removes the temporary pycairo do_install_append() workaround.


WanWizard:
Thanks very much !

